PR #1, opened 2021-07-04
Write guidelines for tutorials

wsh32:
Before starting the git tutorial, I thought it would be a good idea to write a quick outline of what is expected in each tutorial. Open to feedback/ideas

### DIFF
--- a/tutorials/README.md
+++ b/tutorials/README.md
@@ -1,0 +1,29 @@
+# Tutorials
+
+Here you will find interactive tutorials on team tools and processes. To follow
+along, make a new directory with your olin username under the `workspace/`
+directory. There, you can check in your own work to git.
+
+### Contributing
+
+All new tutorials will follow this format:
+
+```
+.
+|-- tutorials/
+|   |-- your_tutorial_name/
+|   |   |-- README.md
+|   |   |-- assets/
+|   |   |   |-- ...
+|   |   |-- workspace/
+|   |   |   |-- template/
+|   |   |   |   |-- ...
+|   |   |   |-- follower_username/
+|   |   |   |   |-- ...
+```
+
+The tutorial content should exist in `README.md`, all data files should exist in
+`assets/`. Include a `workspace/` directory for people following the content to
+work within, and include a `template/` directory within there for people to copy
+to ensure all followers have a starting point.
+


caramulrooney (commented):
Overall, I like this format. This outline is organized so that new members wouldn't get easily confused.
Points of discussion/next steps: 

- What would the format of the templates be? Would they simply be instructions to follow or a worksheet for members to fill out?
- Who is the main audience for these tutorials? Are we going to require them of all (including returning) members?
- We should establish a set of topics for these tutorials before we write any of them. Here's a resource for a bunch of git tutorials: https://www.atlassian.com/git/tutorials

tutorials/README.md
```diff
@@ -0,0 +1,29 @@
+# Tutorials
```
jack-greenberg, line 1: #Use level 2 header `##` and make a single top-level header called something like `# Olin Electric Motorsports Monorepo` or something.

bubbleebubbles:
I like the idea of writing a set of guidelines for the people writing the tutorials (i think that's what you mean by a quick outline of what is expected in each tutorial? correct me if i'm wrong :D) 
Also, yall probably thought about this but do we also wanna do other things besides git like kicad tutorials (ie. how to create a part in kicad)?

bubbleebubbles (approved):
I like this format :)

wsh32:
> Overall, I like this format. This outline is organized so that new members wouldn't get easily confused.
> Points of discussion/next steps:
> 
>     * What would the format of the templates be? Would they simply be instructions to follow or a worksheet for members to fill out?

The idea is that the template provides a starting point for the tutorial. If there is no starting point necessary, it could be just a blank folder. But if all the people following had similar starting points it makes it somewhat easier to automate some checks

>     * Who is the main audience for these tutorials? Are we going to require them of all (including returning) members?

Great question. The idea for the tutorials is to fill the niche that is "team information not found on Google", "people around the team 2+ years can confidently explain", and "all people on the team should know these things". Hopefully we write interactive tutorials that pair with other online resources that reduce the amount of 1 on many teaching that we do.

>     * We should establish a set of topics for these tutorials before we write any of them. Here's a resource for a bunch of git tutorials: https://www.atlassian.com/git/tutorials

Yep! I'll come up with a list soon

PR #2, opened 2021-07-07
Add initial directory structure

jack-greenberg:
This PR only has the top-level folder structure as well as a few
subdirectories for the vehicle folder. As repositories are
integrated, more subdirectories will be create.

The root README has the planned directory structure.


caramulrooney (approved):
Structure & all README.md files look great! I especially like the detail in the libs/README.md and examples/README.md files :)

jack-greenberg:
Let's wait on merging this PR. I might want to add a few more things...

libs/README.md
```diff
@@ -0,0 +1,28 @@
...
+    ├── src_arm/         -> Source files for the ARM library
+    ├── src_avr/         -> Source files for the AVR library
+    ├── include/api.h    -> Common public header
+    ├── test/            -> Unit tests for the library
+    └── BUILD            -> Bazel BUILD file
+```
+
+## To Do
+
+* Is `src_arm` and `src_avr` the best naming pattern? Probably not. Is there
```
wsh32, line 26: src/arm/ and src/avr/ ?

libs/README.md
```diff
@@ -0,0 +1,28 @@
...
+* GPIO library
+
+Libraries will each have their own subdirectory like so:
+
+```
+libs/
+└── example_library
+    ├── src_arm/         -> Source files for the ARM library
+    ├── src_avr/         -> Source files for the AVR library
+    ├── include/api.h    -> Common public header
```
wsh32, line 19: why are we calling it `api.h` rather than `library_name.h'?

libs/README.md
jack-greenberg: Yeah I thought about that too... We'll see. I'll start off by just having files in `src/` and then when we start writing for ARM, we'll figure it out.

libs/README.md
jack-greenberg: Just the pattern I've seen. Then we can do something like

```c
#include "libs/can/api.h"
```

Otherwise it would be

```c
#include "libs/can/can.h
```

which feels redundant. But ¯\_(ツ)_/¯

wsh32: yeah i think 

```
#include "libs/can/can.h"
```

is better because then when you open multiple of the header files at once in an editor youd see

`l/c/api.h`, `l/u/api.h`, `l/s/api.h` which is a lot more confusing than `can.h`, `uart.h`, `spi.h`

awenstrup: yea I think I agree with wesley here. also, I'd like to be able to include our CAN api with `#include "can.h"` if I can.

libs/README.md
```diff
@@ -0,0 +1,28 @@
+# Libraries
```
awenstrup, line 1: How is this folder different from `vehicle/software/mkx/libs`?

vehicle/mkv/README.md
```diff
@@ -0,0 +1 @@
+# MKV
```
awenstrup, line 1: have we given thought to managing multiple vehicles with different branches (like `MKV/production` and `MKV/development`), instead of new directories? both allow us to keep track of old code without losing it, but with multiple branches we can keep the repo smaller, and when starting a new year, we can just checkout a new branch from the previous year instead of copy and pasting the whole vehicle folder.

libs/README.md
awenstrup: Ahhh I see. This is where _shared_ libraries live, while board-specific libraries live in `vehicle/mkx/software/example_board/libs`. 

I think separating out share libraries is smart, but I'd still like to move this closer to the rest of the vehicle software; it feels weird to me to keep this at the root of the directory.

(I've also left a comment below about how we organize code by vehicle, check that out for more context).

Here's what I propose:

```
vehicle
|- hardware
|  |- all_the_hardware_stuff
|- software
   |- shared
   |  |- CAN
   |     |- can.c
   |     |- can.h
   |- example_board
      |- src
      |  |- example_board.c 
      |- libs
         |- all_board_specific_c_and_h_files
```

vehicle/mkv/software/README.md
```diff
@@ -0,0 +1,18 @@
+# Vehicle Software
```
awenstrup, line 1: See above comment on `libs/README.md`

awenstrup (commented):
Most of this looks good to me, super excited to be consolidating a lot of this! Left a few comments, I think there's still more conversation to be had, but this is great. Thanks for taking this on Jack!

libs/README.md
jack-greenberg: hm... yeah we can either have `#include "can.h"` or `#include "full/path/to/can.h"`. Not sure yet which will be easier to implement in Bazel.

libs/README.md
jack-greenberg: But what about projects in the research phase that are using CAN? I don't think we should have cross dependencies between `projects/` and `vehicle/`. I think all _shared_ libraries should be in a common `libs/` folder. And something like `vehicle/software/example_board/libs/` would be a good candidate for something like the LTC BMS chip driver.

vehicle/mkv/README.md
jack-greenberg: @wsh32 thoughts? I think there are a few issues: people won't necessarily think to check branches, and it's a bit more difficult to access things across branches than it is to look in a different folder. 

 I don't think we'll need a MKV/prod and MKV/dev branch, I don't see the added benefit. Also I don't think repo size is going to be an issue.

jack-greenberg: This is definitely a complicated issue. We spent a bunch of time talking about it earlier in the week. Especially for the future where we might want to carry a board over from one car to another, it doesn't make sense to duplicate the code. Here's an idea though...

```
.
└── vehicle/
    ├── hardware/
    │   └── example_board/
    ├── model/
    │   ├── mkv/
    │   │   └── BUILD
    │   └── mkvi/
    └── software/
```

Where the `BUILD` file has something like

```python
vehicle(
    name = "mkv",
    boards = [
        "//vehicle/hardware/example_board",
        ...,
    ],
)
```

So we define a `board` in Bazel, which pulls in a `board`, which pulls in a `cc_binary` which is the application that runs on the board. Thoughts?

libs/README.md
awenstrup: that's a great point about research projects. it still irks me a little bit that libs sits at the root of the repo, but its definitely not as bad as I initially thought. I think I can be convinced, but I'd like to have a conversation about it.

vehicle/mkv/README.md
awenstrup: Good points. I'm still not really sure what the right path here is, I think there are pros and cons to both paths, but I think at this point whatever we do will work (and be better than what we have now). Happy to use the initial structure you proposed.

As for the whole BUILD/Bazel think, that doesn't make a whole lot of sense to me. I honestly don't care (I don't feel a strong personal need to learn Bazel) as long as the interface I interact with is simple.

PR #7, opened 2021-08-19
Adding hitl kicad project

coreyacl:
kicad project files for the hardware in the loop system.


jack-greenberg:
@awenstrup tl;dr it is pretty difficult _at this point_.

The Github action is something I've thought a lot about. It will be possible one day, but with the current version of KiCad, parsing files is still quite difficult. I have played around with the nightly version (KiCad 5.99, which will be released as KiCad 6 sometime in the future), and the file format has changed from the plaintext it currently uses to use S-expression, which is far easier to parse.

Once KiCad 6 is released, and there is a Python API for schematics, it will become much easier and we will definitely have this. Until then, one alternative I've been thinking about is:

In each directory with a `*.pro` file, we will have a folder called `outs/` or `drawings/` or something. There will be files like `schematic.pdf` or `schematic.png` and `layout.whatever`. We can have a Github action that checks will fail if the `*.pro` file was modified more recently than the `*.png` file, which would indicate that the files need to be re-exported.

That wouldn't be too difficult to set up. Happy to chat about it if you believe it would provide value 😄

awenstrup (approved):
Looks good to me! I'd like to get this merged (even if it's not the final schematic) just so we can create the hitl folder :)

PR #14, opened 2021-08-23
Add clang format script to local and CI

jack-greenberg:
Users will have to install clang-format manually with `sudo apt install clang-format`. The script can be invoked to auto format the files using:

```bash
./scripts/clang_format.sh run
```

Also adds a clang_format check to GitHub Actions CI.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,123 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: true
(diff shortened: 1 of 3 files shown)


caramulrooney (approved):
Thanks so much Jack! Looks good. Clang-format doesn't support aligned blocks of one-line #defines and aligned designated initializers, but as long as we keep that in mind, we should be all good :)

jack-greenberg:
I think it does support aligned macros like so

```c
#define FOO    (1)
#define BARBAR (2)
#define BAZ    (FOO)
```

I'm not sure what _aligned designated initializers_ are, but they might be supported?

PR #25, opened 2021-08-25
Add CAN library

jack-greenberg:
Adds CAN library for use in applications. Also includes:

* Legacy CAN library from `MKV-Code` that can be used during the transition so that none of the old code breaks
* A "mock" library that includes all of the API functions to be used in unit tests
* Unit tests that test the receive mock to make sure the receive filters work

The legacy library has a deprecation warning, and should eventually be removed once all the old applications are migrated over.

Should be merged after #13 - Integrate CMocka unit testing library


awenstrup:
Curious if this is happening alongside Hwei Shin and Izzi's CAN work; is there a clear boundary of where this project ends and their's begins? Are the three of you (+ maybe Manu) having integration discussions?

jack-greenberg:
Yeah this is more of a CAN driver that interacts with hardware. Izzi and Hwei Shin’s code is an app-level library that will invoke the functions of the driver. We had a conversation the other night and are synced up on progress :)

The two pieces are independent.

libs/can/api.h
```diff
@@ -0,0 +1,79 @@
...
+#include <stdint.h>
+
+typedef enum {
+    BAUD_250KBPS,
+    BAUD_500KBPS,
+} baud_rate_t;
+
+typedef struct {
+    uint16_t id;
+    uint8_t dlc;
```
awenstrup, line 15: What's a CAN dlc?

awenstrup: Nvm this was very googleable. In case anyone else is curious it's the data length code indicating the number of data bytes in the message.

awenstrup (commented):
looks good to me! let me know if there's anything specific I should look at but the interface seems to make sense :)

jack-greenberg:
Rx and tx both tested with a Canable device and found to be working. Interrupts not yet tested, but would like to merge before anyways--interrupts can come later.

This commit also introduces some Bazel fixes that I missed before having to do with link-time optimization in compiling and dependency issues with linker scripts.

awenstrup:
> Rx and tx both tested with a Canable device and found to be working. Interrupts not yet tested, but would like to merge before anyways--interrupts can come later.
> 
> This commit also introduces some Bazel fixes that I missed before having to do with link-time optimization in compiling and dependency issues with linker scripts.

I can't approve as long as this is still a draft, but I'm happy to approve once that's changed.

awenstrup (approved):

libs/can/can.c
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2021 Olin Electric Motorsports
+ */
+#include "libs/can/api.h"
+#include "libs/can/mob.h"
+#include <avr/io.h>
+
+void can_init(baud_rate_t baud) {
```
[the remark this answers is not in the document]
jack-greenberg: http://wiki.olinelectricmotorsports.com/display/EL21/Git+Revamp+V2

I don't feel like we need to distinguish between public and private things. If a function is private to a specific translation unit (i.e. file), it can be private by just declaring it `static`. Or we can just prefix it with `private_`

PR #32, opened 2021-09-14
Only run github actions on PRs to main branch, not all branches

awenstrup:


jack-greenberg (approved):
lgtm, just curious why this is needed?

awenstrup:
> lgtm, just curious why this is needed?

I noticed this when I put up a PR from awenstrup/feature2 -> awenstrup/feature1. (in that case I could have just pushed the new commits but user1/feature_improvements -> user2/feature is a more realistic use case)

PR #41, opened 2021-09-22
Migrate OEM_Preferred_Parts

jack-greenberg:
This just copies over the parts files from OEM_Preferred_Parts. Work still needs to be done to clean up the files.


coreyacl:
PR looks great. Thanks for sweeping through and dealing with the last steps for this. I pulled the PR and tested the kicad library. Do we want to normalize things like R_ vs r_ for resistors later next week?

bubbleebubbles (approved):

PR #43, opened 2021-09-23
Add KiCad automation

jack-greenberg:
Adds automated KiCad build artifact generation. Built targets are as follows:

* BoM CSV
* Schematic PDF
* Schematic SVG
* PCB SVG (both top and bottom)
* Zip file with GERBERs and Drill files

In CI, these will build, and a comment will be added by @github-actions with images of the schematics and layouts that have changed in the PR, as well as links to all the generated artifacts (which get uploaded to a DigitalOcean storage space that we rent).

Currently, there are a few caveats that I hope to fix one day:

* STEP files don't work because we don't have 3D models for all of our parts. This will be taken care of at some point (cc @coreyacl)
* Schematic SVGs only generate the top-level schematic, so sub-sheets (hierarchical sheets) don't get output as SVGs. They _do_ get output in a PDF though, and the link to the PDF will work
* If a layout doesn't have an edge cut, the build will fail. That means that all boards need an edge cut in order for the build to pass. Not sure how to get around this yet, but I hope to have a solution one day. Also not sure how empty layouts/schematics are handled


bazel/defs.bzl
```diff
@@ -2,6 +2,8 @@ load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
 load("@rules_cc//cc:defs.bzl", "cc_binary")
 
+### cc_firmware
```
mpatil99, line 5: intentional?

mpatil99 (approved):
Looks good to me overall. Really cool to see kibot in action

awenstrup:
Just finished vomiting after looking at the commit history. Otherwise, LGTM; I agree with Manu that `.sh` scripts are hard for me to read, but probably the right way to do what you're doing with them. 

Btw, this is f***in sick! Can't wait to see this in action :)

.github/git-bazel-kicad.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Script to build all KiCad targets that have been changed in the PR
```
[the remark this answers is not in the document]
jack-greenberg: Sounds good!

bazel/defs.bzl
jack-greenberg: Yep!

PR #49, opened 2021-09-27
Updating LV Interface

mpatil99:
Moving pull up resistors for current sensor due to mechanical conflict with BSPD


jack-greenberg (approved):

aditya-sudhakar:
Screen Shot 2021-10-04 at 3 25 55 PM
I don't understand why the pull up resistors are still so close to a connector/where a PCB might be mounted. Like 40% of this board is open space.

PR #50, opened 2021-09-28
Add memory libraries for bootloader

jack-greenberg:


jack-greenberg:
Will be part of different PR.

PR #51, opened 2021-10-01
Add bazel options to build hitl library and run hitl setup script

awenstrup:
Added bazel targets to:

* Build the hitl library from Bazel (the library can also still be installed with `pip install -e .`)
* Run the hitl setup script from bazel

I notably did NOT add the capability to test from Bazel. I looked into it, and saw that bazel doesn't support most of the power features of pytest that I rely on (specifically, the ability to specify a subset of tests from the CLI). I really think testing should be invoked from pytest directly.


jack-greenberg (dismissed):
Looks good to me. Are you using these for development? A good metric for success would be that when you are working on this, you're using Bazel to invoke things.

jack-greenberg (approved):

PR #57, opened 2021-10-05
Add GPIO library for digital IO pins

jack-greenberg:
Adds a new method for defining pins and modifying them to make things easier to mock or stub out in unit testing.

## Usage

Pins can be named by their function and defined by the pin used in hardware. For example:

```c
#include <utils/delay.h>
#include "libs/gpio/api.h"
#include "libs/gpio/pin_defs.h"

gpio_t led0 = PB2;

int main(void) {
    gpio_set_mode(led0, OUTPUT);

    while (1) {
        gpio_toggle_pin(led0);
        _delay_ms(250);
    }
}
```


libs/gpio/pin_defs.h
```diff
@@ -0,0 +1,88 @@
+#include <avr/io.h>
```
awenstrup, line 1: Does this imply an order in which `avr.io.h` and `gpio/api.h` should be included, if including both at the application layer? If so, is that documented somewhere? I could see that being a big "gotcha" down the road, but I'm also not confident this is actually even an issue lol

libs/gpio/mock/gpio.c
```diff
@@ -0,0 +1,36 @@
+#include "libs/gpio/api.h"
```
awenstrup, line 1: What's the point of this if we don't have tests yet? Are tests still a WIP? Do you want to include them in this PR or nah (I don't care either way as long as you're confident this will work)?

PR #62, opened 2021-10-16
CANController Refactor

awenstrup:
This PR transitions the CANController object in the HITL library from an "ECU" based representation of vehicle state to a more direct, CAN signal representation of state. For example:

In the past, to get the throttle position, we might write `harness.throttle.get_state("throttleOut")`

Now, we would instead write `harness.can.get_state("throttleOut")`

This interface makes it more explicit what kind of signal we are asking for, and actually simplifies internal representations a lot, making it easier to edit and maintain this codebase (just look at lines added vs lines deleted).


mpatil99:
LGTM!

mpatil99 (approved):

PR #63, opened 2021-10-22
WIP: can API

hsharriman:


projects/can_api/files/c_templates/generate_can_arr.txt
```diff
@@ -0,0 +1,11 @@
+    //{msg_name} struct initialization
+    {arr_name} = malloc(sizeof(uint8_t) * 8);
```
jack-greenberg, line 2: We shouldn't be using malloc. Use a statically defined array:
```c
uint8_t {arr_name}[{msg_length}];
```

PR #66, opened 2021-10-26
Telemetry Core

mkazc:


jack-greenberg:
Are we going to run into sizing issues with C15 in the LV box? We should check with mechanical...

mkazc:
I don't intend to have a cap that big - the reason for the size of C15 is so that we can test and swap different capacitors, and will have room to place them. So it should be all good :)

jack-greenberg (approved):

PR #68, opened 2021-10-26
Update BSPD/Brakelight code to use new libraries

jack-greenberg:


vehicle/mkv/software/brakelight_bspd/bspd.c
```diff
@@ -0,0 +1,65 @@
+#include "libs/adc/api.h"
+#include "libs/can/api.h"
+#include "libs/timer/api.h"
+#include <avr/interrupt.h>
```
mpatil99, line 4: Hmm do we still need an avr reference here? Feels like this was kinda the point of the lib rewrite?
Actually looking at: https://www.nongnu.org/avr-libc/user-manual/group__avr__interrupts.html, looks sei, cei come from here. 
I wonder if in the future we want to move these to the interrupt library as to remove the need for avr specific imports in app code

vehicle/mkv/software/brakelight_bspd/config.h
```diff
@@ -0,0 +1,51 @@
+#include "libs/adc/api.h"
```
mpatil99, line 1: Nit/question, is worth renaming this file to bspd_config.h purely in an effort to change the line in the BUILD file to smthg more specific?

vehicle/mkv/software/brakelight_bspd/BUILD
```diff
@@ -0,0 +1,19 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+cc_firmware(
+    name = "bspd",
+    srcs = [
+        "bspd.c",
+        "config.h",
```
mpatil99, line 7: With reference to my other comment, this line would become bspd_config.h

vehicle/mkv/software/brakelight_bspd/config.h
```diff
@@ -0,0 +1,51 @@
...
+        .pin_behavior = DISCONNECTED,
+        .interrupt_enable = true,
+        .interrupt_callback = timer0_callback,
+    },
+};
+
+/*
+ * CAN messages
+ */
+#define CAN_MSG_DLC (5)
```
mpatil99, line 44: Cool!

mpatil99 (approved):
Some general nits. Approved, if you decide to ignore/punt them

vehicle/mkv/software/brakelight_bspd/bspd.c
jack-greenberg: Yeah I've thought about this a bit... I think I'm going to keep it as is for now, and we can revisit down the line if we want.

vehicle/mkv/software/brakelight_bspd/config.h
jack-greenberg: Good idea!

mpatil99 (approved):

mpatil99 (approved):

PR #69, opened 2021-11-02
Modify GPIO interrupt implementation

jack-greenberg:
Previously, we had a system where you could register a callback function for _each_ pin on a pin change interrupt vector (like `PCINT0_vect`) so that you could target individual pins, but this added a lot of overhead and slowed down the interrupts substantially and caused errors.

The new implementation changes it so that we have a single callback function for each ISR and we can enable interrupts for a single pin, but the same function gets called. So there's one interrupt per 8 pins, and four interrupts total (PORTB, PORTC, PORTD, and PORTE).

The callback functions are initially declared as `__attribute__((weak))` which means they are "weakly defined", meaning the code will compile but the intention is that if you want to, you can redefine the functions in your code and instead of resolving to the weakly defined function it'll resolve to the one you wrote.

### DIFF
--- a/libs/gpio/api.h
+++ b/libs/gpio/api.h
@@ -10,6 +10,9 @@ typedef struct {
     uint8_t pin;
 } gpio_t;
 
+/*
+ * Set a GPIO pin to be an input or an output
+ */
 #define gpio_set_mode(gpio_pin, mode)                       \
     do {                                                    \
         if (mode == OUTPUT) {                               \
@@ -19,11 +22,35 @@ typedef struct {
         }                                                   \
     } while (0)
 
+/*
+ * Set a gpio pin high
+ */
 #define gpio_set_pin(gpio_pin) (_SFR_IO8(gpio_pin.port) |= (1 << gpio_pin.num))
+
+/*
+ * Set a gpio pin low
+ */
 #define gpio_clear_pin(gpio_pin) \
     (_SFR_IO8(gpio_pin.port) &= ~(1 << gpio_pin.num))
+
+/*
+ * Toggle a gpio pin (high becomes low, low becomes high)
+ */
 #define gpio_toggle_pin(gpio_pin) \
     (_SFR_IO8(gpio_pin.port) ^= (1 << gpio_pin.num))
+
+/*
+ * Get the value of a GPIO pin
+ */
 #define gpio_get_pin(gpio_pin) (_SFR_IO8(gpio_pin.pin) & (1 << gpio_pin.num))
 
+/*
+ * Enable interrupts for the given pin. The interrupt function can then be
+ * defined like so:
+ *
+ * ```
+ * void pcint0_callback(void) {
+ *     // Do interrupt stuff
+ * }
+ */
 void gpio_enable_interrupt(gpio_t pin);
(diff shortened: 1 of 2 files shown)


awenstrup:
The change itself makes sense to me, but I don't think I fully understand the motivation. It seems to me like this change puts more responsibility on the programmer (chiefly, the programmer needs to figure out which pin changed in the interrupt callback, which I don't think was the case before). 

You mentioned the previous implementation was causing errors. How? Is there a way to preserve the ease of programming while resolving those errors?

Also, is there a way to figure out which pins changed without looking at PORTA (or PORTB, or ...) and bit-twiddling? This GPIO library is supposed to prevent us from bit-twiddling, so I want to make sure we aren't undoing any work here.

(the code itself looks great to me, no nits from me)

jack-greenberg:
> The change itself makes sense to me, but I don't think I fully understand the motivation. It seems to me like this change puts more responsibility on the programmer (chiefly, the programmer needs to figure out which pin changed in the interrupt callback, which I don't think was the case before).

True that there is more responsibility on the programmer, but we can't really avoid it. There's no way to see which pin changed without bit twiddling. This is just an unfortunate thing we sorta have to do because of the way pin-change interrupts work (one interrupt for 8 pins).

> You mentioned the previous implementation was causing errors. How? Is there a way to preserve the ease of programming while resolving those errors?

I'm not 100% sure what caused the errors, but the error manifested in interrupts functions being triggered but not behaving correctly, likely because other interrupts occurred or values changed.

> Also, is there a way to figure out which pins changed without looking at PORTA (or PORTB, or ...) and bit-twiddling? This GPIO library is supposed to prevent us from bit-twiddling, so I want to make sure we aren't undoing any work here.

See above. The only workaround would be using the `INT1` and `INT2` pins which are _specific_ external interrupt pins (which you can use to trigger interrupts on rising edges, falling edges, etc), but we aren't using those pins for this board (and there are only two of them for the ATmega16m1.
 
> (the code itself looks great, no nits from me)

:D

awenstrup:
> > The change itself makes sense to me, but I don't think I fully understand the motivation. It seems to me like this change puts more responsibility on the programmer (chiefly, the programmer needs to figure out which pin changed in the interrupt callback, which I don't think was the case before).
> 
> True that there is more responsibility on the programmer, but we can't really avoid it. There's no way to see which pin changed without bit twiddling. This is just an unfortunate thing we sorta have to do because of the way pin-change interrupts work (one interrupt for 8 pins).

Hmmm. I'm thinking about whether or not it's worth providing a helper function like `gpio_t[] get_changed_pins(int port)` or something. Might be helpful, but also adds some computing overhead, and we might run into trouble at high CPU or memory load. Thoughts?

jack-greenberg:
The issue is that we would need some way of storing the values of the pins _before_ the interrupt so that we can see what changed. One idea could be

```c
// gpio.c

static volatile uint8_t pcint0_prev_values = 0;
ISR(PCINT0_vect) {
    pcint0_callback();
    pcint0_prev_values = PINB;
}

// Returns the changed pins between the last interrupt and the current pins
// There will be a 1 wherever a pin changed and a 0 where it didn't
// We could set up this function to have an argument for which interrupt vector you want (pcintN)
void gpio_get_changed_pins(void) {
    return PINB ^ pcint0_prev_values;
}
```

I don't really see this being very useful though. I think it would be easier to just check the values of the pins during an interrupt and do something with them. So whenever an interrupt occurs, always do something with the value of each pin (usually set a byte in a CAN message).

There's an example in the latest BSPD code of how these interrupts work, I'll send it a bit later.

awenstrup:
Just chatted with Jack. The INT0 and INT1 pins give us access to "more powerful" interrupts (that fire only on rising or falling edges, for example), which we can use if we have a use case that requires it. Currently, we don't. Approving.

awenstrup (approved):

PR #72, opened 2021-11-05
Amit kumarh/hitl can sending

amit-kumarh:
This PR adds the ability to manually set signal values on the RKH CAN Bus manually through the set_state function in the CANController object. It also adds a test to test_can.py that uses two CANController objects to verify this functionality.

Also updates the testing .dbc files to scale throttle position from 0 to 100.

### DIFF
--- a/projects/hitl/software/hitl/cancontroller.py
+++ b/projects/hitl/software/hitl/cancontroller.py
@@ -78,15 +78,17 @@ class CANController:
         """
         Set the state of a can signal on the car
 
-        NOTE: This module does not yet support CAN injection, so setting
-        the state of a signal only updates an internal lookup dictionary
         """
         try:
-             msg_name = self.message_of_signal[signal]
+             #find message name
+             msg_name = self.message_of_signal[signal] 
+             #update signals dict to new value and pull out message to send
              self.signals[msg_name][signal] = value
              msg = self.db.get_message_by_name(msg_name)
+             #encode message data and create message
              data = msg.encode(self.signals[msg_name])
              message = can.Message(arbitration_id=msg.frame_id, data=data)
+             #send message
              self.can_bus.send(message)
         except KeyError:
             raise Exception(f"Cannot set state of signal '{signal}'. It wasn't found.")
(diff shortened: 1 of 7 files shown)


projects/hitl/software/hitl/cancontroller.py
```diff
@@ -113,14 +119,16 @@ def _parse_frame(self, msg):
         
         Parses through a CAN frame and updates the self.states dictionary
         """
+        print(msg)
```
awenstrup, line 122: Can we remove these prints before we merge?

projects/hitl/software/hitl/cancontroller.py
```diff
@@ -113,14 +119,16 @@ def _parse_frame(self, msg):
...
         # Get the message name
-        msg_name = self.db.get_message_by_frame_id(msg.arbitration_id)
+        msg_name = self.db.get_message_by_frame_id(msg.arbitration_id).name
         # Decode the data
         data = self.db.decode_message(msg.arbitration_id, msg.data)
+        print(data)
         # Update the state dictionary
         self.signals[msg_name].update(data)
 
-    def __del__(self):
```
awenstrup, line 123: Can we keep the __del__ name? Not sure why it doesn't work yet, but fixing it doesn't need to be in the scope of the PR. If you need to delete a CANController, just manually call can.__del__() for now.

projects/hitl/software/tests/unit/test_can.py
```diff
@@ -44,8 +46,9 @@ def test_create_state_dictionary(can, logger):
...
 @pytest.mark.soft
 @pytest.mark.unit
-def test_set_and_get_state(can, logger):
+def test_set_and_get_state(can, can2, logger):
     logger.info("Testing set_state and get_state...")
 
     can.set_state("ThrottlePos", 50)
-    assert can.get_state("ThrottlePos") == 50
+    time.sleep(.1)
+    assert can2.get_state("ThrottlePos") == pytest.approx(50, rel=0.01)
```
awenstrup, line 54: add a comment explaining the need for pytest.approx here (rounding error)

OR...

We can reduce the resolution of the ThrottlePos message a little bit so 100% = 200, not 255, which gives us much nicer rounding (this would be a .dbc change)

awenstrup (commented):
Overall this looks great! I left a few clean-up notes but nothing big needs to change.

amit-kumarh:
-Removed print statements
-Renamed destructor
-Removed approx statement and changed the scale in both .dbcs for ThrottlePos to .5

jack-greenberg:
Can you update the name of the PR? `Amit kumarh/hitl can sending` isn't very descriptive, but something like `Add manual signal controls to HitL CANController` is better

jack-greenberg (approved):
Looks good.

PR #73, opened 2021-11-07
Create stable voltage ref for strain gauge circuitry.

AntoinetteTan:


AntoinetteTan:
[sensing_controller.pdf](https://github.com/olin-electric-motorsports/olin-electric-motorsports/files/7493407/sensing_controller.pdf)

PR #74, opened 2021-11-09
fix: make AIR+ and AIR- boards rules compliant

elviswolcott:
fix: adjust routing to provide proper TS/GLVS separation of 4mm

fix: reposition components to fix issues with parts on both sides

fix: adjust sizing of relay model

feat: add footprint for resistor


elviswolcott:
I noticed the boards don't have ground planes. I have ideas but would love someone to clarify why we don't need/use a ground plane for these boards.

jack-greenberg (approved):
Thanks for doing this Elvis!

PR #75, opened 2021-11-09
Add service section controller hardware

bubbleebubbles:
Completed Service Section Layout
Spread out the TSAL resistors and added copper pour for heat dissipation. 
Moved TSAL test points away from LV. 
Moved silkscreen to accommodate for the above changes. 

Added mimi silkscreen to the footprints library.


jack-greenberg (approved):
Looks good! Since we shipped this, let's merge the PR.

PR #76, opened 2021-11-11
Add missing BUILD files for MKV hardware

jack-greenberg:


awenstrup:
Assuming the deleted schematic PDFs are because we are auto-generating them now, is that right?

Also, noticing a lot of .pngs and .jpgs being deleted in the sensing folder; why?

Build file and github workflow changes look good to me.

jack-greenberg:
> Assuming the deleted schematic PDFs are because we are auto-generating them now, is that right?

Correct.

> Also, noticing a lot of .pngs and .jpgs being deleted in the sensing folder; why?

They aren't being used, and they're taking up space. They'll always be available in olin-electric-motorsports/MKV_Boards, but we don't need the images in our monorepo since they aren't in use.

awenstrup (approved):
Looks good!

PR #79, opened 2021-12-03
Update module initialization defaults

awenstrup:
Right now, the CANController and IOController don't assume any default values, even when they are initialized. While I'd like to keep their `__init__`s parameterized for flexbility, most people should use the `artifacts/config.ini` file to control these parameters (can channel, can bitrate, etc) most of the time.


mpatil99 (approved):

aramachandran7:
Looks good.

PR #87, opened 2022-01-06
Migrate to KiCad 6

jack-greenberg:
This pull request migrates all KiCad files in the repository to KiCad 6.0 and updates CI to work with the new filetypes.


jack-greenberg:
Currently, there is no support for KiCad 6 continuous integration. We could build our own, but I'm not sure it would be time-effective. If anyone is interested, reach out!

PR #90, opened 2022-01-25
microhitl python wrapper for pyserial

krish-suresh:
PR for arduino serial package for micro hitl


projects/microhitl/ShellIOControl.ino
```diff
@@ -0,0 +1,171 @@
+#include <Shell.h>
+
+void setup()
+{
+  // Prepare serial communication
+  Serial.begin(9600);
+  // Wait after reset or power on...
+  delay(1000);
```
jack-greenberg, line 8: Don't need this delay.

projects/microhitl/pyserialarduino/pyserialarduino/__init__.py
```diff
@@ -0,0 +1,42 @@
+from traceback import print_tb
```
jack-greenberg, line 1: Why do you need this?

jack-greenberg (changes requested):
So it seems like it functionally works, but I'd like to see you simplify the Arduino script a bit. It seems like it's very comprehensive, but a bit hard to follow.

To begin with, I think you can have 3 different "commands" like:

* `gpio_mode`
* `gpio_read`
* `gpio_write`

Then you can have simpler commands like `gpio_mode <PIN> <INPUT|OUTPUT>` or `gpio_write <PIN> <HIGH|LOW>`. This will be good as well because it will separate your huge function into a few different, more readable functions.

For your Python, can you rename the `Arduino` class to `MicroHitl`? It is feels clearer to me. Also, don't print anything in your class functions--libraries shouldn't print things out.

Why is your configure_pin method reading the pin too. It shouldn't do that.

You should be using a Bazel BUILD file, not setup.py. See the `projects/hitl` folder for examples, and ping me if you need help.

Finally, your folder structure is pretty complicated--can you simplify it? Something like

```
/projects/microhitl
    - BUILD
    - microhitl.py -> this is where your class is defined
    - __init__.py -> should be empty
    - microhitl.ino -> the arduino image (this won't be built with Bazel)

**Comment by jack-greenberg** [comment]
Also can you rename the PR? Make it more descriptive and like a statement, the same that you would write for a commit message (like what would you name this patch if it was a single commit?).

**Review by jack-greenberg — CHANGES_REQUESTED** [review]

File: projects/microhitl/ShellIOControl/ShellIOControl.ino [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Function to write data to serial port
+ * Functions to write to physical media should use this prototype
+ * void my_writer_function(char data)
+ */
+void shell_writer(char data) {
+    // Wrapper for Serial.write() method
+    Serial.write(data);
```
jack-greenberg, line 79: Get rid of this line. We don't want to use the default `shell_writer` because it will muck up the things that are returned to the Arduino, like the prompt (`device> `). We should control everything that we print out, not the library.

projects/microhitl/examples/test_read_write.py
```diff
@@ -0,0 +1,11 @@
+from microhitl.microhitl import MicroHitl, Values, PinTypes
```
jack-greenberg, line 1: Can you update this? What would be cool is a loop that blinks an arduino by doing HIGH, sleep, LOW, sleep, and then reads the value of a _different_ pin each loop.

projects/microhitl/microhitl.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def write_pin(self, pin: int, output_value: Values):
+        '''Writes to arduino pin with Value of HIGH or LOW'''
+        self.arduino.write(
+            f"gpio_write {pin} {output_value.value}\r".encode('utf-8'))
+
+    def read_pin(self, pin: int):
+        '''Reads from arduino pin and return true if output contains HIGH'''
+        self.arduino.write(f"gpio_read {pin}\r".encode('utf-8'))
+        return 'read:HIGH' in self.arduino.read(50).decode("utf-8")
```
jack-greenberg, line 29: Use `self.arduino.readline()` instead.

jack-greenberg: (50 is an arbitrary number and reading an entire line (up to `\r\n`) is more conventional.

jack-greenberg (changes requested):

projects/microhitl/microhitl.py
```diff
@@ -0,0 +1,44 @@
...
+class MicroHitl():
+    def __init__(self, port='/dev/ttyACM0', baudrate=9600, timeout=.1):
+        self.arduino = serial.Serial(
+            port=port, baudrate=baudrate, timeout=timeout)
+
+    def write_pin(self, pin: int, output_value: Values):
+        '''Writes to pin with Value of HIGH or LOW'''
+
+        self.arduino.write(
+            f"gpio_write {pin} {output_value.value}\r".encode('utf-8'))
```
jack-greenberg, line 22: Do not encode with `utf-8`. Just call `.encode()`.

jack-greenberg: Also, use `\r\n`, not just `\r`. Do these both for each of the methods.

projects/microhitl/test_read_write.py
```diff
@@ -0,0 +1,18 @@
+from projects.microhitl.microhitl import MicroHitl, Values, PinTypes
+import time
+import os
+
+'''Script to test the read and write of the MicroHitl python wrapper'''
+
+
+if __name__ == "__main__":
+    port = os.getenv("ARDUINO_PORT") or "/dev/ttyACM0"
+    arduino = MicroHitl()
```
jack-greenberg, line 10:
```python
arduino = MicroHitl(port)
time.sleep(2) # Wait 2 seconds for the device to boot up
```

projects/microhitl/BUILD
```diff
@@ -0,0 +1,19 @@
...
+
+py_library(
+    name = "microhitl",
+    srcs = [
+        "microhitl.py",
+    ],
+)
+
+py_binary(
+    name = "microhitl_script",
```
jack-greenberg, line 11: Rename this to just `test_script`.

jack-greenberg (changes requested):

jack-greenberg (approved):
nice work!

PR #93, opened 2022-02-04
Added black python formatter 

krish-suresh:
added black python formatter to lint action runner


jack-greenberg:
Can you update your python formatter PR to only fix files that have changed? Your current PR affects every python file in the repository, but it would be great if it only ran in for the files that changed in the PR. Check out this [Google search for some resources](https://www.google.com/search?channel=fs&client=ubuntu&q=github+actions+black+only+on+changed+files)

PR #98, opened 2022-02-09
Add SPI library

mpatil99:


libs/spi/BUILD
```diff
@@ -0,0 +1,86 @@
...
+#         "test/config328.h",
+#     ],
+#     target_compatible_with = [
+#         "//bazel/constraints:atmega328p",
+#     ],
+#     visibility = ["//visibility:public"],
+#     deps = [
+#         ":spi",
+#     ],
+# )
```
jack-greenberg, line 86: Can you either delete or uncomment this? Or leave a comment above that explains why it's commented out?

libs/spi/legacy/main.c
```diff
@@ -0,0 +1,57 @@
+#include "spi.h"
+#include <avr/interrupt.h>
+#include <avr/io.h>
+#include <stdio.h>
+#include <util/delay.h>
+// #include "spi.c"
```
jack-greenberg, line 6: Remove?

libs/spi/spi.c
```diff
@@ -0,0 +1,127 @@
...
+    gpio_set_mode(spi_state.spi_pin_defs->sck, main_direction);
+    gpio_set_mode(*(spi_state.spi_pin_defs->cs), main_direction);
+
+    // Save mode
+    spi_state.mode = spi_cfg->mode;
+    // Set CS_pin high - active low signal
+    gpio_set_pin(*(spi_state.spi_pin_defs->cs));
+}
+
+void spi_transceive_private(uint8_t txdata, uint8_t* rxdata) {
```
jack-greenberg, line 99: Can you make this `static`? `static void spi_transceive_private(...`

libs/spi/api.h
```diff
@@ -0,0 +1,60 @@
...
+    bool pin_redirect;
+} spi_cfg_s;
+
+/*
+ * Initializes SPI peripheral with the specified parameters
+ */
+void spi_init(spi_cfg_s* cfg);
+
+/*
+ * Transmits byte array and returns data from other party
```
jack-greenberg, line 56: Can you discuss what the args of this function should be? For example, there's no indication that we can pass `NULL` for `rxdata`.

jack-greenberg (changes requested):

jack-greenberg (approved):
Thanks so much for tackling this!

jack-greenberg:
image

Confirmed working.

PR #99, opened 2022-02-11
Add HitL primary board hardware

coreyacl:
Hardware In The Loop Primary Board Rev 1
**Outline page**
https://docs.olinelectricmotorsports.com/doc/hitl-hardware-design-log-zQDYVW9DT6?starred


aramachandran7:
## Layout review first pass 
Were there things you wanted to be reviewed / looked at in particular? 

### Linreg & power
- clever use of Cu planes for the linreg!
- really clever use of Cu planes for the shared 12V net between the `Away` and `Towards` conns!
- CAN- trace to J13 lookin sus
- why do you change trace size on the CAN+ and - lines as you approach J13 and 15? 
- theres a couple ratsnest here and there. drc to find them and fix
- make it clear that the addr jumpers J7-12 are for adjusting addresses in silkscreen, same w AMUX divider jumpers, 
- *You need better jumper silkscreen!* At least label the jumpers for what they're for, you don't need to label the exact pins. 
- label your 0603 0 ohm resistors too, for clarity 

### I2C bus 
- was bus topology intentional, and why? 
- clean routing on sda and scl. 

### I2C level shifter 
- layout looks fine 
- remind me where the 3V3 line comes from? 

### FTDI 
-  3v3 rail comes from an LDO inside here, right? 
- there are no concerns w tying RESET pins to 3v3 that is also comign form this chip

### AMUX 
good 

### ADC 
- label your 0ohm jmp r46, and for other IC's? 
- not a huge fan of how ADCIN goes between R30 and pin, can you shorten / straighten that trace a little? route it under the chip maybe 

### GPIO
- pretty GPIO traces to conn. looks good

### DAC 
- VOUT ss labels on left side of DUT connector are wrong. 
- clever setup with LDAC from FTDI chip.

jack-greenberg:
Maybe add mounting holes so that we can have it sit a bit off the table. That would also allow us to add daughter boards on top of this one and screw them into the mounting holes.

coreyacl:
> Maybe add mounting holes so that we can have it sit a bit off the table. That would also allow us to add daughter boards on top of this one and screw them into the mounting holes.

There are 4 mounting holes already. Are you suggesting we add more?

coreyacl:
> ## Layout review first pass
> Were there things you wanted to be reviewed / looked at in particular?
C: Nothing in particular. Thank you!

> ### Linreg & power
> * why do you change trace size on the CAN+ and - lines as you approach J13 and 15?
C: That was a mistake. I'll fix that

> * theres a couple ratsnest here and there. drc to find them and fix
C: Great catch, ty

> * make it clear that the addr jumpers J7-12 are for adjusting addresses in silkscreen, same w AMUX divider jumpers,
C: 👍 

> * _You need better jumper silkscreen!_ At least label the jumpers for what they're for, you don't need to label the exact pins.
> * label your 0603 0 ohm resistors too, for clarity
C: 👍 
> ### I2C bus
> * was bus topology intentional, and why?
C: Yep! My initial thought is that a bus routing would have fewer reflections, but that may not be justified. 

> ### I2C level shifter
> * remind me where the 3V3 line comes from?
C: The FTDI chip has an internal 3V3 regulator. 

> ### FTDI
> * 3v3 rail comes from an LDO inside here, right?
C: Yep

> ### ADC
> * not a huge fan of how ADCIN goes between R30 and pin, can you shorten / straighten that trace a little? route it under the chip maybe
C: 👍  

> ### DAC
> * VOUT ss labels on left side of DUT connector are wrong.
C: Well that's embarrassing. Fixed

Thanks for the feedback!

jack-greenberg:
Ah sorry, I was referencing the automated PR comment when looking at layout. It hasn't updated in a bit (CI won't run if you have merge conflicts). I'll take a look locally.

jack-greenberg (approved):
Let's GOOOO

PR #105, opened 2022-02-15
Integrate LTC6811 libraries

jack-greenberg:
This commit will integrate the LTC681x (general) and LTC6811 (specific) libraries for use in the BMS and LVBMS code.


aramachandran7 (commented):
Making sure I understand: 
`bms_hardware.c` and `.h` are the only files you wrote and allow the ltc libs from online to use our SPI and GPIO libs. You didn't make any changes to the ltc libs from online, correct?

jack-greenberg:
Mostly. I had to make a few modifications to the LTC6811 libraries because they used dynamic memory allocation which we don't have, and there were a few other weird syntax things.

aramachandran7:
So did you only make changes to `ltc681x` or to the `ltc6811` files too? Also curious about desired directory structure - since the lvbms will have to use the `ltc681x` header too (the lvbms uses the ltc6810 chip), does it make sense to place the `ltc681x` headers in the general libs directory?

jack-greenberg:
> So did you only make changes to `ltc681x` or to the `ltc6811` files too? Also curious about desired directory structure - since the lvbms will have to use the `ltc681x` header too (the lvbms uses the ltc6810 chip), does it make sense to place the `ltc681x` headers in the general libs directory?  

I had to make changes to both, but they were mostly just changing function names and changing dynamic allocation to stack allocation. 

For dir structure, I thought about that, but `//libs` doesn't feel right either. Let's keep this for now and maybe down the line we can find a better place for it. The library is public so your code will still be able to import it.

aramachandran7 (approved):

PR #107, opened 2022-02-15
Add BMS code

jack-greenberg:
Add in MKV BMS code.


aramachandran7:
Wakeup commands, ADC polling, PEC error counting all is well documented & makes sense. 

`pack_voltage += ICS[ic].stat.stat_codes[0]` - does `...stat_codes[0]` contain the state of charge voltage from the `adcvsc`? If so, does `...stat_codes[0]` get written to in the `rdcv` command? If yes, make this clear in your comments.   

`uv` and `ov` bit operations in code are super clean. I'll take your word on if odd and even bits store UV and OV flags respectively, and that you can access the ST register in the way you did. 

Nice work!! Looks super clean. :D

aramachandran7:
Other points for `voltage_task` function, after writing some lvbms code (for the LTC6810 Linear Tech BMS chip): 
- may need to call the ltc6811_rdstat command in the voltage task function in your bms code, to read back the status registers
- Also, check your OV and UV settings. Looks like you divided 4.2V by 100uV but not 16 as well to get 42000 instead of 2625
- Looks like you also need to multiply your sum of all cells measurement by 20 and 100uV to get a usable floating point voltage (or maybe just by 1uV to get a value in mV so that it can remain as a uint16_t)
- 
image
image

jack-greenberg:
I think it'd be a good idea to split this into a few different PRs. I'll probably do that soon. PRs would probably be:

* Add 64m1 support for bazel
* Fixes to SPI library
* BMS code

mpatil99 (approved):
per converstation

jack-greenberg:
Remaining issues:
* PWM for fans
* Test charging
* CAN message regularity (timing functions)
* BMS temperature min/max reporting
* Test current sensor
* Setting thresholds for temperature and voltage

PR #110, opened 2022-02-18
Inertial measurement unit board for gokart

drewnotdrew:
Complete board schematic and begin board layout for IMU.


elviswolcott:
All these commits were just me fighting with CI and cleaning up the rescue libraries that KiCAD loves to curse us with.

elviswolcott:
Overall this is a **really** clean layout. It looks great!
The form factor, organization, and labelling are great.

I noticed a few minor things, nothing we shouldn't be able to fix in the meeting tonight.

---

Here you can see a few places where you ended up with acute angles, mostly along 5V. You should be able to replace these with 90 degree angles.
Screenshot from 2022-03-31 15-45-54

---

This is the only thing I'm seeing that is a real issue is the crystal placement:
Screenshot from 2022-03-31 15-39-57
It's important to have the crystal as close to the micro as possible. We can look at ways to shift things around minimally to move it up. If we need more room we could go ahead and swap the buck converter for an 5V LDO which would give us more space.

---

Here, it should be possible to swap CS to pin 17 to running SCK through a via. Not a big deal but will leave your ground plane slightly cleaner.
Screenshot from 2022-03-31 15-38-48

---

Another thing that falls under small tweaks that could go eitheir way is moving the pull up resistor on RESET down so that the SPI lines run under it.

Screenshot from 2022-03-31 17-20-09
With the resistor here we can get rid of the vias.

jack-greenberg (changes requested):
* (Git hygiene) There are a ton of files you should delete from this PR. All the .DXF, .SLDPRT, .SVG files, as well as both of the `.svg2shenzhen-cache` folders. Also remove all references to the template board
* Add DNP silkscreen for your CAN termination resistor (R2)
* Add silkscreen for LEDs in the buck converter for 12V (the LED that turns on with 12V) and 5V (for the LED that turns on with 5V
* Other than that, looks great!!

jack-greenberg:
Also, please make sure to run

```shell
./scripts/install_kicad_git_filters.sh
```

From the root of your repository and re-add the changed files. This will help us avoid some merge conflict stuff.

jack-greenberg (approved):
Shipped, so approved.

PR #111, opened 2022-02-20
Add fixes and updates to CAN API

jack-greenberg:
This splits up the original `can_receive_xxx()` functions in the CAN API into two separate `can_receive_xxx()` and `can_poll_receive_xxx()` functions, which is more inline with the original CAN library.

Also adds in support for CAN interrupts, for completeness's sake.


mpatil99 (approved):
Looks good. Changes make sense.

PR #115, opened 2022-02-20
Python linter Black added to CI to only check changed files

krish-suresh:


jack-greenberg (approved):
Assuming this works, looks good to me. Let's merge.

PR #116, opened 2022-02-24
Go Kart BMS Core

elviswolcott:


elviswolcott:
Not sure why this has artifacts for the service section controller...

coreyacl:
**Schematic comments**:
* R5 and R6 should both be 1k. This has to match the Ibias and Icmp resistors on the peripherals which have 1k for both. I can explain what this does in person if you're interested. 
* Add a 1uF bypass cap to U4
* Slight addition to the schematic. Could you add a circuit that looks like this: https://docs.olinelectricmotorsports.com/doc/cooling-fan-testing-OvJZhh16Zy
    * I did some fan testing this weekend which confirmed a couple things, including what voltage the fans want their PWM signal. 
* Add a 2nd main connector with 4 pins which includes {GND, 12V, CAN+,CAN-} so that this board can pass those signals onto the next board. 
* Use a right angle microfit for these connector. 
* UF (ultrafit) for J3 is overkill. Please use a microfit. 

**Layout Comments:**
* Where did you get the DXF from? Looks nice!
* On the car, we usually follow HV spacing rules for anything that comes close to the cells, like the iso-spi lines on J2. Since the segments aren't technically HV in this case I think it's fine, I'll just ask that you give the connector a tad more breathing room, and make sure there aren't any copper planes in the vicinity. (Like GND)
* I think you're using an old transformer, not sure if we have any more of those lying around, but you can check. Or switch the part out to the newer one, T_LAN_SM91501ALE
* It's a _little wonky_ to have the main board connector be in the center. Could you move this to the side of the board along with the 2nd main connector you'll be adding?
    * You should be able to shift the Lin-reg circuitry down and move the connectors up 
image
* We have board space, and especially for power you want the circuitry to have lots of low-impedance paths to GND. Add more vias to GND here in the power section.
image
* It's tempting to add vias to pads, but it's generally more expensive to manufacture and not necessary in this case since we have a lot of board space. Move all vias from pads away from them by at least a via width.  
* Since the lin-reg (U3) will be generating heat, let's make it easier for it to dissipate that heat by adding a copper pour around it. You can add one for 5V on the right and GND on the left. 
* Great job keeping a clean GND plane

Great first pass! Slack me if you have any questions about the feedback. Excited to see the next revision :D

jack-greenberg:
Looks great! Since we shipped, let's merge. I'll approve the PR.

jack-greenberg (approved):

PR #118, opened 2022-02-27
Fix issues with template buck converter

elviswolcott:
fix: update footprint for 100H inductor
fix: update layout to match schematic of buck converter


jack-greenberg (approved):
Benefit of the doubt :)

PR #123, opened 2022-03-08
Add updated MKV throttle code

mpatil99:
roast it pls


vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
+    bool flag_implausibility;
+    uint16_t throttle0_mapped;
+    uint16_t throttle1_mapped;
+    uint16_t throttle_values[8];
+    uint16_t rolling_sum;
+    uint8_t implausibility_counter;
+    bool brake_pressed;
+    bool ready_to_drive;
+    air_state_e air_state;
+} throttle_state_s;
```
jack-greenberg, line 54: I don't think a typedef is needed here. Maybe just something like

```c
struct throttle_state {
    ...
    ...
} throttle_state = { 0 };
```

This will also create the variable throttle state and initialize everything to zero.

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
+// future
+typedef enum {
+    INIT = 0,
+    IDLE,
+    SHUTDOWN_CIRCUIT_CLOSED,
+    PRECHARGE,
+    TS_ACTIVE,
+    DISCHARGE,
+    FAULT,
+} air_state_e;
```
jack-greenberg, line 40: Prefix with `AIR_STATE_`

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
+
+#define ROLLING_AVG_SIZE          8
+#define IMPLAUSIBILITY_TIMEOUT_MS 100
+
+uint8_t deviation_counter = 0; // keeps track of elapsed time of >= 10%
+                               // deviation in mapped throttle values
+
+// volatile uint8_t gFlag = 0x00; // Global Flag. Tells us that gFlag will
+// change and to assign it to an unsigned 8bit integer volatile bool send_can =
+// false;
```
jack-greenberg, line 22: Multiline comment.

vehicle/mkv/software/throttle/throttle.c
jack-greenberg: I feel luke-warm about having a single struct to contain all state, buuuut I don't think it's a huge deal? I think you should add comments for each member of the struct about what it is though, unless it's obvious. Also, group like-things together, like flag_implausibility and implausibility_counter.

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
-	temp /= ROLLING_AVG_SIZE;
-	gThrottleOut = temp;
+void pcint0_isr(void) {
+    // Check if shutdown circuit node states
+    throttle.e_stop_sense = !gpio_get_pin(SS_ESTOP);
+    throttle.inertia_switch_sense = !gpio_get_pin(SS_IS);
+    throttle.bots_sense = !gpio_get_pin(SS_BOTS);
 }
 
+uint16_t mapThrottleVoltage(throttle_s* throttle) {
```
jack-greenberg, line 81: snake case plz `map_throttle_voltage`

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
+
+            if (throttle_state.send_can) {
+                can_send_throttle();
+                can_send_m192_command_message();
+            }
+        }
+    }
 }
+
+// can timeout?
```
jack-greenberg, line 256: ?

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,256 @@
...
+                throttle_state.brake_pressed = brakelight.brake_gate;
+                can_receive_brakelight();
+            }
+
+            if (can_poll_receive_dashboard() == 0) {
+                throttle_state.ready_to_drive = dashboard.ready_to_drive;
+                can_receive_dashboard();
+            }
+
+            switch (throttle_state.state) {
```
jack-greenberg, line 206: Manu. You need `break` statements :|

jack-greenberg (changes requested):
Delete `throttle_refactored.c` and move `throttle_old.c` to a `legacy` folder.

Maybe we should move the can_api fixes to a diff PR?

vehicle/mkv/software/throttle/throttle_config.h
```diff
@@ -0,0 +1,70 @@
...
+const throttle_potentiometer_s throttle_l = {
+    .adc_pin = ADC9,
+    .throttle_min = THROTTLE_L_MIN_COUNTS,
+    .throttle_max = THROTTLE_L_MAX_COUNTS,
+};
+
+/*
+ * Timer config
+ */
+void timer0_isr(void);
```
awenstrup, line 58: why are we setting up the timer if it's ISR is empty? Are we using this to time CAN messages?

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,276 @@
...
+                // With no implausibility...
+                throttle_state.implausibility_fault_counter = 0;
+
+                // NOTE: If we decide to do a non-linear map, that would go here
+                int16_t torque_request = pos_min * TORQUE_REQUEST_SCALE;
+
+                SET_TORQUE_REQUEST(torque_request);
+            }
+
+            if (throttle_state.send_can) {
```
awenstrup, line 270: I assume these will send at the freq specified in the yaml (100 Hz), and not at the same speed as the main loop (1000 Hz), right?

vehicle/mkv/software/throttle/utils/timer.c
```diff
@@ -0,0 +1,33 @@
...
+
+void timer1_isr(void);
+
+timer_cfg_s timer1_cfg = {
+    .timer = TIMER1,
+    .timer1_mode = TIMER1_MODE_CTC,
+    .prescalar = CLKIO_DIV_1,
+    .channel_a = {
+        .channel = CHANNEL_A,
+        .output_compare_match = 4000,
```
awenstrup, line 16: I'm a little out of my league on this one, but I can't remember if we run our ATMEGAs at 16 MHz or 12 MHz. I think they can run up to 16, but I think the grey bins indicate that our crystal is 12... I really don't know either way, but if you're also unsure, I can look into it.

awenstrup (commented):
Yo this code looks soooo good. I left a few comments, but as far as I can tell it's really close.

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,276 @@
...
+#define IMPLAUSIBILITY_TIMEOUT_MS (100)
+#define MOTOR_ANTICLOCKWISE       (1)
 
-void storeThrottle(void) {
-	readPots();
-	mapThrottle();
-	checkPlausibility();
+// Set this to lower than 10 to limit torque request (i.e. 5 limits to half of
+// max torque)
+#define TORQUE_REQUEST_SCALE (10)
```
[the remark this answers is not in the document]
jack-greenberg: Ah, no actually. The torque request is scaled by .1, so if we request 255, the actual request is 25.5 N*m.

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,276 @@
...
+    gpio_enable_interrupt(SS_ESTOP);
+    gpio_enable_interrupt(SS_IS);
+    gpio_enable_interrupt(SS_BOTS);
+
+    sei();
+
+    // Initialize motor controller direction
+    m192_command_message.direction_command = MOTOR_ANTICLOCKWISE;
+
+    can_send_m192_command_message();
```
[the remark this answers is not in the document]
jack-greenberg: DOPE. This is super good to know, thanks Alex!

vehicle/mkv/software/throttle/throttle_config.h
jack-greenberg: This is just the function prototype, so that we can pass the reference to the struct (basically lets us pass a pointer to the function in the struct), and then we actually _define_ the function in `throttle.c`.

The `timer0_isr` sets a variable called `send_can = true` and triggers every ~10ms, so yeah 100Hz.

vehicle/mkv/software/throttle/utils/timer.c
jack-greenberg: Ah, so I actually did make a mistake here. Freq of the clock is 16MHz. There's a prescalar that drops it down, and then the output compare match further divides this. So to get a 1ms tick timer, we do

16000000 / 8 (prescalar) = 2000000 Hz
2000000 / 2000 = 1000 Hz

vehicle/mkv/software/throttle/throttle.c
jack-greenberg: Yeah this will be set at 100Hz.

vehicle/mkv/software/throttle/utils/timer.c
```diff
@@ -0,0 +1,33 @@
+#include "libs/timer/api.h"
+#include <avr/interrupt.h>
+
+volatile uint32_t time;
+volatile bool run_1ms;
+volatile uint32_t start_time;
+
+void timer1_isr(void);
+
+timer_cfg_s timer1_cfg = {
```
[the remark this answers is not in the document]
jack-greenberg: This is a function prototype, not an empty function.

vehicle/mkv/software/throttle/throttle.c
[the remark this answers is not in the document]
jack-greenberg: Technically I never set it to 1, so we didn't have lockout, but also the inverter would never be enabled haha. I will set it so that as soon as we have ready-to-drive, we enable the inverter.

vehicle/mkv/software/throttle/throttle.c
```diff
@@ -1,738 +1,276 @@
...
-
-	sprintf(uart_buf, "v2: %d", gThrottle2Voltage);
-	LOG_println(uart_buf, strlen(uart_buf));
+    // Voltage range between 0% and 100% pedal travel
+    int16_t range = throttle->throttle_max - throttle->throttle_min;
 
-	sprintf(uart_buf, " ");
-	LOG_println(uart_buf, strlen(uart_buf));
-}
+    // Pedal position percentage (between 0 and 1)
```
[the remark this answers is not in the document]
jack-greenberg: Good point. This code actually won't work, so I modified it a bit. Good catch!

jack-greenberg (dismissed):

vehicle/mkv/software/throttle/throttle.c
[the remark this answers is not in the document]
awenstrup: Ah, yea. So this version of the code will max out at 25.5 Nm (probably smart for just spinning wheels in the air). In that case, I might ask how we plan to scale up our torque request to max motor torque (230 or 240, I can't remember). Do we add another coefficient to map to our motors torque range, or do we just increase this scalar by a factor of ~9? Or am I just missing something?

vehicle/mkv/software/throttle/utils/timer.c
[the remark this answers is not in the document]
awenstrup: I was talking about the timer_cfg. Unless I know the 16M1 like the back of my hand, I can't determine the intention behind the timer_cfg values just by looking at this.

jack-greenberg:
Blocked by merging BMS.

jack-greenberg (approved):
Mostly complete. Last few bugs will be ironed out on different branches.

PR #124, opened 2022-03-09
Arturo/sensing controller layout

ArturoJoya:
A submission of the sensing controller PCB layout


ArturoJoya:
The actual PCB work was done in sensing_controller1 (the project and the file)

PR #125, opened 2022-03-17
CAN csv to yaml decoding tool.

mpatil99:
Converts mkv csv format to new can yaml.

With a little more work, can be made to convert an arbitrary format to can csv yaml. With a tiny bit more work, can convert CAN yaml format to any arbitrary format.


jack-greenberg:
Closing because we no longer need this functionality (the BMS CAN message structure has changed).

PR #126, opened 2022-03-24
Updates to accumulator interface for cooling fans

jack-greenberg:


jack-greenberg:
idk why the service section layout had changes. We can revert them before we merge.

PR #127, opened 2022-03-25
Add firmware test board

MetaKor:


jack-greenberg (changes requested):
# Feedback Round 1

_In no specific order..._

- Don't put so many traces under the ATMEGA (ideally zero), and don't put _any_ traces under the crystal or CAN transceiver
- Screw pin terminal is backwards (see 3d model)
- Move crystal closer to ATMEGA, same with capacitors and resistors
  - For layout inspiration for this, look at other boards like the service section or BSPD
- Every GND should have it's own via, don't just share them around
- Don't put traces underneath the crystal
- Use fewer bottom traces unless just going from through-hole to through-hole (like jumpers to connector)
  - You've created cut-outs in your ground plane, which is a bad idea (I can explain why in person)
- Move programming header closer to edge of board and so that 3-pin side is flush with the board
- Add mounting holes on all four corners
- For connector, no need to connect grounds together since you have a ground plane
- Try to not put 12V through via, and move the lin-reg circuitry closer to the 12V supply
- Add some silk screen please, like for pins on the jumpers and label the CAN screw-pin terminals, and label the CAN jumper
- Move references (i.e. R12, J3) around so they are easier to see
- Add radius (smoothing) to corners of board
- Big connector should be fully flush with the board edge
- Flip the board horizontally so that the connector is on the left
- Clearly label pin 1 on the big connector (currently cut off)
- Move things around better. Your placement I think is hindering your routing. Like move the ATMEGA closer to the middle of the board so that you can route out to the left and then above or below

coreyacl:
**Schematic:**
* Consider using an ADC pin (35-40 inclusive) from the hitl to measure the 5V rail. I've found it useful to start a test with "does this board even have 5v?"
* Also consider breaking out the DAC pin from the ATmeaga to one of the ADC pins as well. 
* Consider adding more comments. I can inuit what's going on, but there's room for their to be MORE clarity.
* Consider using the CAN pins on the hitl instead of braking out CAN independently. You can definitely do both for covering bases.
* You should add a pull-down resistor to the FET for reset (Q1). If you don't, then the gate will be floating which may or may not randomly pull the chip reset LOW. No bueno 
* Let's chat about potentially adding current limiting resistors to the GPIO pins, or if even that's worth it. 
* I think it's cool/neat how you placed all pins of same register next to each other on the 44pin connector. Nice touch. 

**Layout**
* First glance, looks _nice_ 👌 
* Yes, CAN is robust, but I think it's good to rehearse best practices for routing differential signals. Kicad helps with that by going to "route" and clicking on "differential pair". Try using this tool for most of the run and feel free to use regular traces for when you can no longer make the differential routing work. 
* No other comments come to mind. Great job!

jack-greenberg:
Most changes look good to me. Is the pull-down on the FET Q1 in the correct spot? 
image

I'm just not sure haha.

Also, didn't see anything about current-limiting resistors for GPIOs or tuning the track length for the CAN differential pairs. Did you get a chance to chat with @coreyacl about this?

jack-greenberg:
Last few things!
* Let's talk about adding 100ohm resistors between all of the GPIO pins 
* Make sure you run the setup script for the repo: `./scripts/install_kicad_git_filters.sh` and then add in all the changes
* Run differential routing on your CAN traces. We can chat about this during the meeting.

jack-greenberg (approved):
Looks great Luke! Thank you so much for working on this! We will probably ship out Friday with a batch of other boards.

PR #128, opened 2022-03-30
Add support for CCLS linting in code editor

jack-greenberg:
I may be the only person to use this, but if anyone wants to set up linting using the CCLS language server (I think you can use it in VS Code and neovim), feel free!


BUILD.bazel
```diff
@@ -14,3 +15,10 @@ buildifier(
     mode = "check",
     tags = ["manual"],
 )
+
+refresh_compile_commands(
+    name = "refresh_compile_commands",
+    targets = {
+        "//vehicle/mkv/software/...": "--platforms=//bazel/platforms:atmega16m1_avr -c opt",
```
mpatil99, line 22: is it worth adding 64m1 while were here?

mpatil99 (approved):
looks good, just one question

PR #129, opened 2022-03-31
Add Go-Kart Throttle Schematic + Layout

amit-kumarh:
Whoops not me realizing I never opened a PR for this...

Here's the go-kart throttle.


jack-greenberg (approved):
Since we shipped, I'll approve and we can merge.

PR #135, opened 2022-04-13
Milestone: MKV Wheels Spinning

jack-greenberg:
This is a PR to track work being done for Wheels Spinning. Once we have wheels spinning, we'll merge and tag the repository. This is for all the loose ends and random one-off things that need to get done.

## BMS Changes
Previously, we got all BMS temperatures at the same time. This took a very long time because we had to wait for SPI and I2C. Now, the temperature_task function doesn't loop through the muxes and channels.

Instead, each time the function is called, it increments the channel it is using and performs the measurement for all the LTC6811s. Then, the next time it is called, it switches to the next channel. So each time the function temperature_task is called, we only measure one temperature, but it increments each time.

This should fix timing issues in the BMS that was causing the state machine to take 500+ ms (!!!) to run.

Fixed PWM so that the fans will are now controllable.

This also calibrates under and overtemperature thresholds for the BMS (-20C to 60C).

## AIR Control Changes
* Added a 100ms delay before checking for AIRs welded in discharge because there's a slight delay between the AIRs opening up that we need to account for
* Added a 4.5sec delay in `init` before checking the status of the IMU because of the startup hardware delay on the service section controller

## Throttle Changes
* Calibrated throttle pedals
* Throttle code had inverter disabled, fixed that
* Changed scale of torque request
* Changed direction of motor

## BSPD/Brakelight
* Now drives cooling pump LSD when we do ready2drive

## Hardware
* AIR Mount boards rev 2 (new form factor)

### Service section rev 2
 * changed 12V IMD net to 5V to make layout changes easier, 5V IMD is supplied after 5V AIR control fuse
 * Power Main: Added larger test points for +Batt and GND for easier testing
 * IMD Latch: changed 12V nets to 5V, swapped inverting and non-inverting inputs f
or both op amps, removed resistor divider at IMD_STATUS_OUTPUT that goes to the ATm
ega
 * TSAL/AIL: changed 12V_TSAL to a 5V input, changed ts+ side resistor dividers to 1k and 120 (bottom of the op amp), changed AIL logic to use an nmos
 * Power Secondary: removed HV lin reg, added 15V - 5V lin reg and added a fuse fo
r the 5V output. 400-15V lin reg can be used to provide 15V OR we can change resist
ors to provide 5V and short the output of lin reg to the output of the buck.
 * Micros: added a programming header for isospi for easier firmware testing, adde
d test points for CANH and CANL for firmware debugging.
 * Updated IMD Latch resistor dividers.
 * Changed 5V lin reg capacitors to 2 1uF capacitors.
 * Added a protection diode for lin reg.
 * Added a 0 Ohm jumper between 5V tsal and 15V buck output in case the buck conve
rter can output 5V and we no longer need the lin reg.
 * Spread out the tsal resistors to help with heat dissipation.
 * Changed IMD connector to a RA footprint (same footprint as VT tho)
 * Changed Shutdown connector to an 8-pin RA microfit
 * Changed TS connector to a 4 pin VT microfit
 * shifted BMS connector slighly
 * changed bms and air control programming led silkscreen labels to general,___ fault
 * removed 2-pin RA connector used for final shutdown
 * added footprints for RA 8-pin and 6-pin microfit to formula footprint library
 * shifted ts connector up (prev location was difficult to use)
 * added small copper plane around lin reg for heat dissipation
 * changed 12V IMD test point silkscreen to 5V IMD
 * removed FAN buffer
 * Shifted C902 left due to clearance issues with emeter.
 * changed IMD resistor divider so that the op amp can handle when IMD outputs 16V.
 * changed clearance of the plane around tsal resistors to 30 mil.

### LV Interface

 * Added cooling pump low-side drive to interface board to be driven by BSPD board


vehicle/mkv/software/air_control/utils/utils.c
```diff
@@ -44,7 +44,7 @@ int get_bms_voltage(int16_t* voltage) {
         }
     } while (rc != 0);
 
-    *voltage = bms_core.pack_voltage;
+    *voltage = bms_core.pack_voltage << 4;
```
jack-greenberg, line 47: TODO: This needs to be `<< 8`

amit-kumarh (approved):

ayushchakra (approved):

elviswolcott (approved):

mkazc (approved):

PR #137, opened 2022-04-18
feat: comp ready AIR mounts

elviswolcott:
feat: spec out new relays rated for 200W

feat: move resistors off board to allow series configuration and larger heatsinks

feat: follow HV spacing for uncoated board

feat: clearly label HV and GLV regions on PCB

feat: use micromatch connectors only

feat: use consistent pin ordering to prevent risk of swapped connectors

docs: clearly label schematic regions and spacing rules

fix: make discharge circuit connect to TS instead of BAT

fix: swap flipped redcubes or AIR-

fix: misc schematic changes


jack-greenberg:
Don't forget to update the `BUILD` file to get it to build properly.

coreyacl:
At first glance, the boards looks _really clean_, well done! Also great job keeping HV spacing so that we don't have to conformal coat, excellent. 

Notes:
Weird naming thing going on. Air MINUS has `TS+` and `BAT+` and the Air PLUS project has all of the `TS-` and `BAT-` net labels. Am I misunderstanding something? 
Air plus
* Could you make a note that H1 and H2 are just holes for soldering wire to for discharge resistors. 
* The NC relay you chose looks good. I'm concerned about the 15V max operating voltage. Our battery goes up to 16V so we should test to see if that will be an issue.
* I believe your discharge circuitry is incorrect. Discharge is supposed to discharge the leftover energy in the motor controller capacitors when the shutdown circuit OPENs, and as you have it wired now the discharge relay will not discharge anything if its connected between TS+ and bat+. Instead it should go from TS+ (motor controller+) to TS- through a connector. 
I'll continue my review tomorrow...

elviswolcott:
Overall only 2 minor changes on the plus side layout and schematic. Mainly focused on cleanup to make it all more readable going forwards (and useful as a resource in the future). If I understand correctly we can just connect the AIR+ to GND because there is no LSD so I got ride of COIL-, if this is correct I'll also go ahead and swap the 6pin RA to a 4pin VT for J2.

I have 2 outstanding questions:
1. Right now the emeter connection is on the AIR- (fused TS-), I'm not sure what the emeter connection is exactly used for so I'm not clear whether we should be swapping that over to TS+ like in the ESF schematic. Same thing applied to BSPD_current_sensor.
2. On the AIR+ there is un-fused TS+ going into the fuse connector. I still haven't figured out what this is for and it seems like it could be nasty if that shorts. If we aren't using it for something I'd like to remove it.

I also added the 12 ohm resistor as suggested because it doesn't seem worth the hassle of trying to test/burning out a $$ relay. The coil resistance and max rated voltage are the same on both relays so I added it to both.

elviswolcott:
Switched to 4pin microfit VT where possible. Removed unfused TS connection and made pin ordering a bit more consistent across the boards.

jack-greenberg:
No clue why `lint` is failing in CI. I will investigate.

jack-greenberg:
Rebased to main to pick up changed to CI that will let your PR pass CI.

jack-greenberg (approved):
Nice job! Going to approve the PR since we are about to ship.

PR #139, opened 2022-04-20
dchadiuk/service section controller v2

bubbleebubbles:
see the last 2 commits for the changes made or view the outline page here: https://docs.olinelectricmotorsports.com/doc/service-section-controller-debugging-nWrbFqsE0T


coreyacl:
Great work!
A couple notes:
* C704 is technically a 16V cap, we should make sure to use a higher rated cap for that. 30V+ would be ideal to give the cap some wiggle room and derate suffeciently. NO need to make a schematic change, just a note for populating. 
* The lin-reg (https://www.digikey.com/en/products/detail/diodes-incorporated/AP7370-50W5-7/10235827) has an 18V input MAX. we might consider adding an over-voltage protection diode after the fuse to try and protect the IC in case the buck acts up and output higher than 15V. The 18V zener diode we use on the regular buck should work!
* U701 is the 15V->5V linreg. IT could definitely use more copper to dissipate the heat it will generate. Just a bit around it should suffice, nothing too dramatic 
image
* I am, however, concerned about the power dissipation of voltage divider resistors. Could we use the space in the top right area here for placing these resistors? 
image
* Not sure how we chose our inductor but if we're looking to slim down the physical size, coilcraft.com is a pretty nice place look for inductors.
No further comments, great job!

aramachandran7:
Nice work, great documentation!
Notes: 
- Is there any chance we want to run the HV buck at 5V out instead of 15V out? and resultingly ignore the AP7370 linreg? should we note this in the schem somewhere + add the fuse / 0Ohm jumper in between the `+15V_FUSED` and `+5V_TSAL` net? 
- Drop u801 and connect `FAN_PWM_RAW` net directly to `FAN_PWM`. We no longer us the schmitt trigger here. 
-  Good move to provide +12V SHDN to IMD conn J301. 
- R305 308 310 no longer divide to 1.63V. Either change resistor values or the schematic text. 
- Add schem text for why R312 exists. 
- Add a truth table / enunciate the IMD logic with a picture in the `imd_latch` schem page. 
I'm sure there's small things I'm missing, but solid job! :D Good shit dude

aramachandran7:
Nice work again. Mostly nitpick at this point. 
Notes:
- Could you note in the schem that R601/604/606/607 each dissipate 0.0432W at TS+=400V and a Trise estimate to justify the massive Cu plane? 
- Drop u801 and connect FAN_PWM_RAW net directly to FAN_PWM. We no longer us the schmitt trigger here.
- ^^ same comment from last time did you forget? :shrug: either this or add a 0 ohm jumper to short the output if you want to keep the schmitt trigger footprint just in case? 
- why is F204 after F202? is there a reason for this? 
- Now that we've tested that the HV Buck can output 5V without issue and know that we want to use it in our final system, is it correct to say that we **don't** want to make any schematic changes or notes to indicate that we may run the HV Buck output at +5V and ignore u701? That seems sort of counterintuitive. Thoughts? 
- for u701 have you done thermal calcs to know if you need a little copper plane so this dude doenst get super hot? 
- TP205 label on layout is wrong
- seriously no DRC errors? impressive, you should have seen my STEP PCB :grimacing:

bubbleebubbles:
responding to adi's comments (numbering them in order of bullet points): 
1. Will add the note!
2. oops i did forget about the Fan PWM raw, will add. 
3. n/a
4. F204 is before F202 because we need to get 5V to the 12V IMD net somehow but the 5V output of the buck can't be easily attached to the fuse without a lof of vias. easier thing was to extend the 5V Air control plane and separate the two with a fuse. 
5. I think we can keep it, that's why the 0 ohm jumper is in between. We can populate buck only and then if we ever need lin reg, its there. 
6. fok, i forgot to add the copper plane for thermals. good thing you noticed. 
7. will change label. 
8. hehe I got rid of all my DRC errors before pushing ;)

bubbleebubbles:
i realized that some of the AIL lights we got from digikey use 12V or 15V so keepin lin reg is actually more desirable in this situation.

jack-greenberg:
There are files that need deleting and we need to rebase your PR. I can do this once you're done making changes or I can walk you through it, your call.

bubbleebubbles:
ye i realized after pushing that i accidentally put a bunch of the fabrication files into the pr. plz help

coreyacl:
thanks again for taking the time to test so thoroughly and make changes 💯 

**Schematic:**
* Main page -> Change Rev1 to Rev2
* Shutdown -> Great text comments!
* IMD Latch -> The IMD input voltage divider outputs 6.7V when the LV battery is fully charged at 16V. 
    * If we use the current chip, the absolute ratings suggest that we're fine 5V above VCC. 
* Do we have these comparators in stock? 
* Did you simulate the RC timer? Not required, but prolly a good idea 
* TSAL/AIL -> We may have to reduce the resistance of R610 since the opto-isolator requires around 16mA to operate ON. We can switch out the resistor when we're populating it.

**Layout**
* Looks _cleaaaan_
* Had we more time , I think it'd be great to be more careful with HV tracing. I know there aren't explicit rules on HV spacing on the HV side but we should be conscious of physical requirements.
* At full pack voltage there will be 200V between the plane and the trace entering here. No bueno. 
* Let's increase the clearance of the copper planes to 30mil for good measure. 
image
image

No other comments. Great job! :+1:

PR #140, opened 2022-04-22
Fixes while testing throttle

amit-kumarh:
Calibrating throttle potentiometers, fixed state setting


vehicle/mkv/software/throttle/throttle.c
```diff
@@ -16,22 +16,6 @@
 #include <math.h>
 #include <util/delay.h>
 
-#define IMPLAUSIBILITY_TIMEOUT_MS (1000)
```
jack-greenberg, line 19: Moved `#defines` (macros) to `throttle_config.h`

PR #141, opened 2022-04-22
Modify BMS temperature to use asynchronous architecture

jack-greenberg:
Previously, we got all BMS temperatures at the same time. This took a very long time because we had to wait for SPI _and_ I2C. Now, the `temperature_task` function doesn't loop through the muxes and channels.

Instead, each time the function is called, it increments the channel it is using and performs the measurement for all the LTC6811s. Then, the next time it is called, it switches to the next channel. So each time the function `temperature_task` is called, we only measure one temperature, but it increments each time.

This should fix timing issues in the BMS that was causing the state machine to take 500+ ms (!!!) to run.

This also calibrates under and overtemperature thresholds for the BMS (-20C to 60C).

### DIFF
--- a/vehicle/mkv/software/bms/bms.c
+++ b/vehicle/mkv/software/bms/bms.c
@@ -24,6 +24,7 @@ enum State {
 };
 
 enum AIR_State {
+    AIR_STATE_INIT,
     AIR_STATE_IDLE,
     AIR_STATE_SHUTDOWN_CIRCUIT_CLOSED,
     AIR_STATE_PRECHARGE,
@@ -213,7 +214,7 @@ static void state_machine_run(void) {
                 bms_core.bms_state = FAULT;
             }
 
-            if (air_state != AIR_STATE_IDLE) {
+            if ((air_state != AIR_STATE_IDLE) && (air_state != AIR_STATE_INIT) && (air_state != AIR_STATE_FAULT)) {
                 bms_core.bms_state = DISCHARGING;
             }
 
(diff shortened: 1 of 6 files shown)


vehicle/mkv/software/bms/bms.c
```diff
@@ -203,21 +157,25 @@ static void state_machine_run(void) {
      */
     uint32_t ot = 0;
     uint32_t ut = 0;
+    int16_t min_temp, max_temp;
+
+    // uint16_t pack_temperature;
+    rc = temperature_task(&ot, &ut, &min_temp, &max_temp);
+    // bms_core.pack_temperature = pack_temperature;
+
+    // if (ut > MAX_EXTRANEOUS_TEMPERATURES) {
```
jack-greenberg, line 166: Need to uncomment/figure out what's going on. Might be that we just need to _actually_ calculate the thresholds.

vehicle/mkv/software/bms/bms.yml
```diff
@@ -40,10 +40,17 @@ publish:
...
           offset: 0
           scale: 0.0001 * 256
-      - name: pack_temperature
-        slice: 24 + 16
+      - name: min_temperature
+        slice: 24 + 8
         unit:
-          type: int16_t
+          type: uint8_t
+          name: degC
```
jack-greenberg, line 47: Not actually degC. We need to figure out if we can linearize, or if we need to just report as voltages.

PR #142, opened 2022-04-22
testing LVBMS SPI!

aramachandran7:
Updated SPI lib to support peripheral A; however `spi_write_read()` inside ltc681x.c `read_68()`  cmd blocks `lvbms.c` superloop. 
Problem likely lies inside `spi_write_read` and integration with peripheral A in `spi.c` SPI lib. Requires a little more digging into uC SPI peripheral.


coreyacl:
Great work!
These are mostly nit-picks so I'll be explicit if I think there is a MUST HAVE change required before shipping the board.
Schematic:
* Nice LT1716 part, but it doesn't have an MFG number. Please add MFG, manufacturer part number to make it easy to export this part in the BOM.
* The footprint is a bit jank with overlapping text but I don't think this _needs_ to be fixed.
* Are there any single comparator packages for this part? I can't find a pinout...
* Have you confirmed that you can program the chip using `PROG_A`?
* For sake of professionalism and your portfolio, please remove text above your name 😬 
Layout
* Looks great! Love the testpoint additions. no comments here

aramachandran7:
Sweet thx
1. done
2. unclear which footprint you're referencing, ignored 
3. page 2 of [ds,](https://www.analog.com/media/en/technical-documentation/data-sheets/LT1716.pdf) we're good. 
4. yep thats how we always do it, reference ss mega, it checks out
5. forgot to remove that text :rofl: I thought it was funny the first time around ... 
6. (layout) - thanks. will clean up ss a little before ship tn.

jack-greenberg:
Split this PR into two? One for hardware that we can merge into `rc/competition` and one for firmware?

PR #147, opened 2022-05-01
Adi r/lvbms fw restart

aramachandran7:
we’re getting appropriate SPI MISO response from the 6810 for `LTC6810_adcv()` and `LTC6810_pollAdc()` functions, but for `LTC6810_rdcv()` & `LTC681x_rdcv()` we aren’t getting either MOSI or MISO activity. 

@mpatil99 and i verified it wasn’t an issue in `bms_hardware.c` as both `SPI_read_write()` and SPI_write_array()` functions in there result in MOSI activity, so it must have been how we end up calling `LTC681x_rdcv_reg()`

i added in the address bits to the first command byte for `LTC681x_rdcv_reg()` function, but the problem we’re seeing now (no MOSI activity) would happen irregardless of what we have set as our address bits.

### DIFF
--- a/vehicle/mkv/software/lvbms/lvbms.c
+++ b/vehicle/mkv/software/lvbms/lvbms.c
@@ -1,0 +1,50 @@
+#include "lvbms_config.h"
+#include "libs/adc/api.h"
+#include "libs/gpio/api.h"
+#include "libs/timer/api.h"
+#include "libs/spi/api.h"
+#include "vehicle/mkv/software/lvbms/can_api.h" // TODO: why?? 
+#include "avr/interrupt.h"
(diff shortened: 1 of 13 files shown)


vehicle/mkv/software/lvbms/lvbms.c
```diff
@@ -0,0 +1,67 @@
+#include "lvbms_config.h"
+#include "libs/adc/api.h"
+#include "libs/gpio/api.h"
+#include "libs/timer/api.h"
+#include "libs/spi/api.h"
+#include "vehicle/mkv/software/lvbms/can_api.h" // TODO: why??
```
jack-greenberg, line 6: This is so that you can access the functions defined in the CAN API like `can_send_lvbms`.

PR #149, opened 2022-05-03
Add CAN viewer dashboard application

jack-greenberg:
A picture is worth 1000 words:

image

Easily extensible, can report any live CAN data. Main limitation is that we can't do graphs, but that's what we have Grafana for.

### DIFF
--- a/projects/canviewer/canserver.py
+++ b/projects/canviewer/canserver.py
@@ -1,0 +1,40 @@
+import cantools
+import can
+import threading
+
+db = None
+
+
+def listener_fn(can_bus, callback):
+    """Thread that runs all the time to listen to CAN messages
+
+    References:
+      - https://python-can.readthedocs.io/en/master/interfaces/socketcan.html
+      - https://python-can.readthedocs.io/en/master/
+    """
+    while True:
+        msg = can_bus.recv(1)  # 1 second receive timeout
+        if msg:
+            callback(msg, db)
+
+
+def init_can(channel, bitrate, rx_callback, dbc, filters):
+    global db
+    can_bus = can.interface.Bus(
+        channel=channel,
+        bustype="socketcan",
+        bitrate=bitrate,
+    )
+
+    listener = threading.Thread(
+        target=listener_fn,
+        name="listener",
+        kwargs={"can_bus": can_bus, "callback": rx_callback},
+    )
+
+    listener.daemon = True
+    listener.start()
+
+    db = cantools.database.load_file(dbc)
+
+    return can_bus, db
--- a/projects/canviewer/canviewer.py
+++ b/projects/canviewer/canviewer.py
@@ -1,0 +1,163 @@
+from canserver import init_can
+from ui import render_dashboard
+
+import argparse
+from os import listdir
+import time
+import signal
+import sys
+import numpy as np
+
+from rich.layout import Layout
+from rich.live import Live
+
+db = None
+
+FILTERS = [
+    {
+        "can_id": 0x000,
+        "can_mask": 0x7E0,
+        "extended": False,
+    },
+    {
+        "can_id": 0x0A0,
+        "can_mask": 0x7FF,
+        "extended": False,
+    },
+]
+
+SHUTDOWN_NODES = {
+    "ss_bspd": None,  # BSPD
+    "ss_is": None,  # THROTTLE
+    "ss_bots": None,  # THROTTLE
+    "ss_estop_driver": None,  # THROTTLE
+    "ss_hvd": None,  # AIR
+    "ss_hvd_conn": None,  # AIR
+    "ss_mpc": None,  # AIR
+    "ss_bms": None,  # AIR
+    "ss_imd": None,  # AIR
+    "ss_tsms": None,  # AIR
+}
+
+VEHICLE_VALUES = {}
+
+VEHICLE_STATES = {
+    "AIR Control": [None, None],
+    "BMS": [None, None],
+    "Throttle": [None, None],
+}
+
+
+def sigint(signal_received, frame):
+    exit(0)
+
+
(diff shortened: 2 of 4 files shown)


projects/canviewer/canserver.py
```diff
@@ -0,0 +1,40 @@
...
+      - https://python-can.readthedocs.io/en/master/interfaces/socketcan.html
+      - https://python-can.readthedocs.io/en/master/
+    """
+    while True:
+        msg = can_bus.recv(1)  # 1 second receive timeout
+        if msg:
+            callback(msg, db)
+
+
+def init_can(channel, bitrate, rx_callback, dbc, filters):
```
amit-kumarh, line 21: Are the filters used anywhere? If not can we remove them?

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,163 @@
+from canserver import init_can
+from ui import render_dashboard
+
+import argparse
+from os import listdir
```
amit-kumarh, line 5: Is this import used? (also sys)

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,163 @@
...
+    thermistor_R1 = (Vout * R1) / (Vin - Vout)
+    temperature = 1 / ((np.log(thermistor_R1 / R2) / beta) + (1 / T2)) - 273.15
+
+    return temperature
+
+
+def get_val(signal, data):
+    val = data.get(signal)
+
+    if val != None:
```
amit-kumarh, line 70: minor: could write as `return val if val is not None`, (and/or could remove the else clause), although this might be more legible. Either works

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,163 @@
...
+        return str(val)
+    else:
+        return None
+
+
+def get_ss_val(signal, data):
+    val = data.get(signal)
+
+    if val:
```
amit-kumarh, line 79: Is this outer `if` (and the else statement at the end of this function) necessary, given that if val is neither "OPEN" nor "CLOSED" we will return `None` anyway?

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,163 @@
...
+            return None
+
+
+def rx_callback(msg, db):
+    global SHUTDOWN_NODES
+    global VEHICLE_VALUES
+
+    try:
+        message = db.decode_message(msg.arbitration_id, msg.data)
+    except Exception as e:
```
amit-kumarh, line 94: Do we know what exception is thrown here? I trust that this works but it would be cleaner if we could specify that

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,163 @@
...
+        return
+
+    if msg.arbitration_id == 0x00B:  # BSPD
```
amit-kumarh, line 97: Don't know if you're planning on maintaining this, but if all the signals are unique, can we loop through all of the signals in a message, and if the key is in SHUTDOWN_NODES, VEHICLE_VALUES, or VEHICLE_STATES, then set it's value, something like:

```
for signal_name in message:
    if signal_name in SHUTDOWN_NODES:
        SHUTDOWN_NODES[signal_name] = get_ss_val(signal_name, message)
    if signal_name in VEHICLE_VALUES:
...
```
I think it would clean up the code and make it more easily expandable since then we can just add/remove signal names in the constant table dictionaries without worrying about changing code down here. Could keep ID differentiation just for the states if we wanted to since their structure is a bit more complicated (and in either case if we keep the IDs can we put them in a constant dictionary up top so they're easily referenced/changed?)

projects/canviewer/canserver.py
amit-kumarh, line 21: Are we using the filters anywhere? If not can we remove them?

projects/canviewer/canviewer.py
amit-kumarh, line 70: minor: can either drop the else or rewrite as `return str(val) if val is not None`. Might be more legible this way though

projects/canviewer/canviewer.py
amit-kumarh, line 79: minor: Do we need this outer if (or the else statement at the end), given that the function will return None if val is not "OPEN" nor "CLOSED" anyway?

projects/canviewer/canviewer.py
amit-kumarh, line 97: If all the signals are unique, can we loop through all of the signals in a message, and if the key is in SHUTDOWN_NODES, VEHICLE_VALUES, or VEHICLE_STATES, then set it's value, something like:

```
for signal_name in message:
    if signal_name in SHUTDOWN_NODES:
        SHUTDOWN_NODES[signal_name] = get_ss_val(signal_name, message)
    if signal_name in VEHICLE_VALUES:
...
```
I think it would clean up the code and make it more easily expandable since then we can just add/remove signal names in the constant table dictionaries without worrying about changing code down here. Could keep ID differentiation just for the states if we wanted to since their structure is a bit more complicated (and in either case if we keep the IDs can we put them in a constant dictionary up top so they're easily referenced/changed?)

projects/canviewer/canviewer.py
amit-kumarh, line 5: minor: Is this import (and sys on line 8) used?

projects/canviewer/canviewer.py
amit-kumarh, line 94: Do we know what exception we're catchign here? I trust that this works but it would be cleaner if we could specify that

amit-kumarh (commented):
Bit late and not sure if you're planning on maintaining this but had a bit of dead time and figured I'd leave some comments here. I'm also happy to develop this, would be a fun side project and I think a program like this would be really useful, particularly in making CAN more accessible without needing to learn cantools.

PR #153, opened 2022-07-14
Add power distribution library [TUTORIAL]

amit-kumarh:
Add power distribution library to monorepo (firmware week 1 tutorial)
- Support for Dashboard, service section, and throttle

__DO NOT MERGE__

### DIFF
--- a/tutorials/power_distribution/BUILD
+++ b/tutorials/power_distribution/BUILD
@@ -1,0 +1,16 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+cc_library(
+    name="power_distribution",
+    srcs = [
+        "power_distribution.c",
+    ],
+    hdrs = [
+        "api.h",
+    ],
+    deps = [
+        "//libs/gpio",
+    ],
+)
+
+# bazel build --config=16m1 -c opt //tutorials/power_distribution
--- a/tutorials/power_distribution/BUILD
+++ b/tutorials/power_distribution/BUILD
@@ -1,5 +1,3 @@
-load("//bazel:defs.bzl", "cc_firmware")
-
 cc_library(
     name = "power_distribution",
     srcs = [
--- a/tutorials/power_distribution/BUILD
+++ b/tutorials/power_distribution/BUILD
@@ -1,7 +1,7 @@
 load("//bazel:defs.bzl", "cc_firmware")
 
 cc_library(
-    name="power_distribution",
+    name = "power_distribution",
     srcs = [
         "power_distribution.c",
     ],
--- a/tutorials/power_distribution/api.h
+++ b/tutorials/power_distribution/api.h
@@ -7,20 +7,26 @@ gpio_t THROTTLE_POWER = PB2;
 
 /*
  * Power on the given device
+ * Args:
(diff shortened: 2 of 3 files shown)


tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,16 @@
...
+    srcs = [
+        "power_distribution.c",
+    ],
+    hdrs = [
+        "api.h",
+    ],
+    deps = [
+        "//libs/gpio",
+    ],
+)
```
jack-greenberg, line 14: 👍

tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,16 @@
...
+
+# bazel build --config=16m1 -c opt //tutorials/power_distribution
```
jack-greenberg, line 16: 👍

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,26 @@
...
+
+/*
+ * Power on the given device
+ */
+void power_device(gpio_t pin);
+
+/*
+ * Power on all devices
+ */
+void power_all();
```
jack-greenberg, line 16: I think this is fine, but usually if a function doesn't take arguments, I specify `void`:

```c
void power_all(void);
```

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,26 @@
...
+#include "libs/gpio/pin_defs.h"
+
+gpio_t DASHBOARD_POWER = PB0;
+gpio_t SERVICE_SECTION_POWER = PB1;
+gpio_t THROTTLE_POWER = PB2;
+
+/*
+ * Power on the given device
+ */
+void power_device(gpio_t pin);
```
jack-greenberg, line 11: I wouldn't name the argument `pin`, I'd probably name it `device` since that is what it refers to. I would also add a description in the comment about what the argument _is_. The idea is that someone should be able to read the docstring and understand what the function does and how to use it, and part of that is understanding what the arguments mean.

tutorials/power_distribution/power_distribution.c
```diff
@@ -0,0 +1,21 @@
+#include "tutorials/power_distribution/api.h"
```
jack-greenberg, line 1: :+1:

jack-greenberg (changes requested):
Looks good! Just a couple minor things.

tutorials/power_distribution/power_distribution.c
```diff
@@ -0,0 +1,21 @@
...
+
+void power_device(gpio_t device) {
+    gpio_set_pin(device);
+}
+
+void power_all() {
```
jack-greenberg, line 7: Technically this should also have `void` as the argument. Same for `power_off_all`.

tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,16 @@
+load("//bazel:defs.bzl", "cc_firmware")
```
jack-greenberg, line 1: Technically not needed because this isn't a firmware binary.

PR #154, opened 2022-07-16
Add CAN viewer GUI

amit-kumarh:
This PR is an expansion on PR #149, @jack-greenberg's terminal CAN dashboard. I chose to separate it to leave the TUI variant on its own should we choose to use it. 

2022-07-16-114846_1858x909_escrotum

### Feature list
- Qt GUI
- Allows for YAML based config, which means adding and removing signals requires no code changes - just a single line of YAML
- Native support for both Socketcan and our new Seeedstudio CAN dongles through a command line argument, elliminating the need for a bridge script when using this tool

#### Revision 2
Simplified YAML config file by changing the Vehicle States document from a list representation of the component state/fault to a dictionary with the signal name as the key and the state or fault as the value. 

This makes reading those values to display them slightly more cumbersome (and importantly, NOT compatible with Python < 3.7), but provides the benefit that the entire program is now __message agnostic__. There is no hard-coding of message names anywhere and the entire dashboard can be configured without touching the Python code and just changing the YAML

#### Revision 3
- Made processing functions fully configurable - they can now be defined in the utils file and configured in the YAML with no additional code changes required. See the updated README for more info
- Improved docstrings and comments
- Added a kill flag when the application is exited to allow the CAN thread to close gracefully
- Renamed throttle state message from `state` to `throttle_state`


jack-greenberg:
This is... SICK! One thought: it would be cool if you used the DBC file as a dependency, and then for your YAML file you could just write which signal from the DBC you wanted to use as `<message>.<signal>`. So for instance, if you wanted to see the `SS_TSMS` signal, you could just put in your YAML file:

```yml
column_1:
- air_control_critical.ss_tsms
```

Does that make sense? I believe you can use cantools for DBC parsing, and you can use the DBC file as a `data` dependency in Bazel:

image

```python
py_binary(
    ...
    data = ["//vehicle/mkv:mkv.dbc"]
    ...
)
```

amit-kumarh:
I do agree that the YAML as it is is a bit messy. And the DBC is actually already used, that's how the CAN messages are parsed. I'm still a bit unsure how specifying the full `message.signal_name` gives any more information though.

As it is the first two columns can be changed with just the name of the signal, and we can refactor the states config by 
doing something like this and adjusting the way the GUI gets the data from the dictionary, which eliminates the messages document, and using the DBC to pull the message ID for the top level keys, which eliminates the message IDs document

```
air_control:
- air_state:
- air_fault:
```

Something like this would remove a good chunk of the hard-coding from the states portion of the rx_callback, abstracting more away from the code and into the config while simplifying the config itself. How does that look?

amit-kumarh:
no rush, whenever you get a chance. See top level comment for change list

projects/canviewer/canserver.py
```diff
@@ -0,0 +1,40 @@
+import cantools
```
jack-greenberg, line 1: Add a comment about the purpose of this file. Why are we creating a thread? What is it doing?

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,117 @@
...
+VEHICLE_STATE_SIGNALS = [
+    signal for message in VEHICLE_STATES.values() for signal in message.keys()
+]
+
+
+def sigint(signal_received, frame):
+    exit(0)
+
+
+def convertVtoT(x, Vin=3, R1=10000, R2=100000, T2=348.15, beta=3988):
```
jack-greenberg, line 35: You should prob make a separate utils.py file that contains utilities that we use, including this function. It's not directly responsible for the CANviewer

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,117 @@
...
+        VEHICLE_VALUES,
+        VEHICLE_STATES,
+    ) = yaml.safe_load_all(config_file)
+# Messages that correspond to vehicle states/faults
+VEHICLE_STATE_SIGNALS = [
+    signal for message in VEHICLE_STATES.values() for signal in message.keys()
+]
+
+
+def sigint(signal_received, frame):
```
jack-greenberg, line 31: Can you figure out a way to gracefully close the CAN thread from this too?

jack-greenberg: Currently, all this does is change the exit code when you press <C-c>

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,117 @@
...
+    try:
+        message = db.decode_message(msg.arbitration_id, msg.data)
+    except Exception as e:
+        return
+
+    for signal_name in message:
+        if signal_name in SHUTDOWN_NODES:
+            SHUTDOWN_NODES[signal_name] = get_val(signal_name, message)
+        elif signal_name in VEHICLE_VALUES:
+            if signal_name in ["max_temperature", "min_temperature"]:
```
jack-greenberg, line 67: I know i wrote this, but i dislike it lmao. it would be cool if there was a way to move the logic for processing a signal elsewhere, like maybe you can put the function name in the YAML file, and then you just see if the signal has a processing function, and if so, you call it. Make sense? I can clarify offline.

projects/canviewer/config.yml
```diff
@@ -0,0 +1,31 @@
...
+air_p_status:
+air_n_status:
+ready_to_drive:
+pack_voltage:
+max_temperature:
+min_temperature:
+D1_DC_Bus_Voltage:
+---
+# Vehicle States (state and fault)
+air_control_critical:
```
jack-greenberg, line 24: Why are these organized by message name, where the previous signals aren't?

jack-greenberg (changes requested):

projects/canviewer/canviewer.py
amit-kumarh: Right now the signals are key value pairs in the YAML, with the key being the name and the value being whatever the CAN thread gives us. We could
- add a separate document to the YAML with key value pairs of signal names and processing functions if they require one. It's the path of least resistance, and leaves a simple config document, but it does require adding another document to the YAML
- Change the structure of the YAML so the values are no longer lists
```
---
# Vehicle Values
max_temperature: [~, processing_function]
throttle_pos: [~]
```
Then the main processor would have to try to get the element at index 1. If it's there, run it through the appropriate processor, if not, send it straight back to eliminate some of the YAML verbosity.

This would not require an extra document, but it would add in that list into our config, which no longer allows the really simple `signal_name:` configuration.

Thoughts?

projects/canviewer/config.yml
amit-kumarh: The message is used both by the GUI (so it can display the message name on the left column of the states table) and by the CAN receiver (so it knows which message a signal comes from).

Theoretically we could eliminate the need for it if:
- We made a dictionary for all of the signals and their corresponding messages from the DBC, eliminating the second functionality
- We added some restructuring between the dictionary that we get after filling in CAN values and the one we send to the GUI so it knows where to display all the values (so for example it knows to place `air_state` in the AIR control row, or for that matter, even knows the AIR control row exists)

In my opinion, that's more work than it would be worth, and syntactically it does make some sense that you specify that you want the state/fault from a certain component as opposed to the others where the source of the message isn't as relevant (eg  it would be irrelevant to the user that ss_bots comes from the BSPD board, but saying you want the state signal from the throttle board is justifiable)

projects/canviewer/canviewer.py
jack-greenberg: Why can't you do

```yaml
---
max_temperature:
  processing_function: convert_v_to_t
throttle_pos:
something_else:
```

And then in code you can check to see if the keys have any values, and if so, look for something called `processing_function`?

I am in favor of keeping everything in a single config file.

projects/canviewer/config.yml
jack-greenberg: Ah I think I see. I guess what I mean is that the system should be unified. In other words, you should either have to specify the message name _and_ the signal name, or just the signal name, not a mixture of both. Ya dig?

Also, I think best practice is to not have duplicate signal names, so every signal name should be unique. If it isn't (like maybe I mixed two things up), we should rename them, like from `state` to `throttle_state` or something like that. You can do this in this PR if you want or I can do it in a separate one.

projects/canviewer/canviewer.py
amit-kumarh: Well it would be in the same config file, just another YAML document. In any case, I'll do some testing but yeah this is a much better solution :+1:

projects/canviewer/config.yml
amit-kumarh: If we keep standardization of state and fault names, how's this for a solution - we can just specify the message name, and use the DBC to pull the names of the signals. Something like this:
```
#Vehicle States/Faults
air_control_critical:
bms:
throttle:
```
This (to me at least) makes logical sense because the user can specify they want the state and fault from say the throttle board, which is more pertinent than what the actual signal name is. Not technically consistent, but it is much simpler.

We could also do this in reverse, where we take the signal names, and use the DBC to find what the names of the messages are.

Essentially, the dictionary that gets passed into the GUI would be the same, we would just need to do some leg work with the DBC at the start of the program to fill in the gaps and build what is specified in the YAML now.

As far as uniqueness of signal names - they are technically unique right now, although I agree that just calling something `state` is terrible practice - pretty sure that was me when I wrote the throttle YAML. I'll change that in the next rev.

projects/canviewer/canserver.py
richardli03: Echoing this! From the perspective of someone who doesn't really have a good background on the way the system works, it'd be really nice to have explanations for individual scripts as well as the readme for the whole system.

projects/canviewer/canviewer.py
```diff
@@ -0,0 +1,105 @@
+"""
+Callback function when receiving a CAN message, launch and update GUI
+"""
+from canserver import init_can
+from gui.gui import Window
+from utils import *
```
amit-kumarh, line 6: Programming heresy with the wildcard import, I know, but it's a file we control and allows the user who's adding a processing function to just define it in utils and set in in the YAML without having to touch this code

jack-greenberg (approved):
Awesome job! I just tested it locally and it looks sweet.

PR #156, opened 2022-07-17
Power Distribution Tutorial

ayushchakra:


tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,18 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+
+
+cc_library(
```
jack-greenberg, line 5: I usually prefer listing `srcs` before `hdrs`.

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,14 @@
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+gpio_t DASHBOARD_POWER = PB0;
+gpio_t SERVICE_SECTION_POWER = PB1;
+gpio_t THROTTLE_POWER = PB2;
+
+void power_device(gpio_t pin);
+
+void power_all();
```
jack-greenberg, line 10: If there are no arguments, you should specify `void` as the argument:

```c
void power_all(void)
```

Same below and in `power_distribution.c`.

tutorials/power_distribution/power_distribution.c
```diff
@@ -0,0 +1,26 @@
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+gpio_t DASHBOARD_POWER = PB0;
+gpio_t SERVICE_SECTION_POWER = PB1;
+gpio_t THROTTLE_POWER = PB2;
+
+void power_device(gpio_t pin) {
```
jack-greenberg, line 8: I wouldn't call it `pin`. Mostly because they represent devices, so I would probably call it `device` or something like that.

jack-greenberg (changes requested):
Looks good, just a couple small things! And be sure to lint your code by running

```sh
$ ./scripts/clang-format.sh reformat
$ bazel run //:buildifier
```

tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,18 @@
+load("//bazel:defs.bzl", "cc_firmware")
```
jack-greenberg, line 1: Technically not needed because this is a library, not a firmware image.

PR #157, opened 2022-07-18
Power Distribution Tutorial

elviswolcott:


tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,30 @@
...
+/* Power ON a device
+ *
+ * warning: board may exhibit unexpected behavior if DEVICE is not
+ * DASHBOARD_POWER, SERVICE_SECTION_POWER, or THROTTLE_POWER
+ */
+void power_device(gpio_t device);
+
+/* Power ON all devices
+ */
+void power_all();
```
jack-greenberg, line 19: It's generally good practice to use `void` as an argument for functions that don't take arguments.

```
void power_all(void);
```
Same below and in your `.c` file

jack-greenberg (changes requested):

tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,20 @@
+load("//bazel:defs.bzl", "cc_firmware")
```
jack-greenberg, line 1: Technically not needed because you are not using `cc_firmware`.

PR #158, opened 2022-07-18
power distribution firmware learning assignment 1

caramulrooney:


tutorials/power_distribution/BUILD
```diff
@@ -0,0 +1,16 @@
+load("//bazel:defs:bzl", "cc_firmware")
```
jack-greenberg, line 1: Your import statement should be `//bazel:defs.bzl`. There should only be one `:`, and it's to separate the folder from the file.

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,18 @@
...
+#include "libs/gpio/pin_defs.h"
+
+gpio_t DASHBOARD_POWER = PB0;
+gpio_t SERVICE_SECTION_POWER = PB1;
+gpio_t THROTTLE_POWER = PB2;
+
+void power_device(gpio_t device);
+// Turn on the specified device
+
+void power_all();
```
jack-greenberg, line 11: Generally speaking, if a function doesn't take arguments, you should put `void` between the parentheses:

```c
void power_all(void);
```

Same below and in other file.

jack-greenberg (changes requested):
Nicely done! Just a couple small things to get it to build, and some aesthetic changes.

tutorials/power_distribution/BUILD
jack-greenberg: Actually though, you don't need to import `cc_firmware` because you are not creating a firmware binary--you are creating a C library.

PR #159, opened 2022-07-18
Complete power distribution library tutorial

mkazc:


tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,33 @@
...
+/*
+Set power GPIO pin high.
+*/
+void power_device(gpio_t gpio_pin);
+
+/*
+Set dashboard power, service section power, and
+throttle power GPIO pins high.
+*/
+void power_all();
```
jack-greenberg, line 22: Generally if there are no arguments, you can just put `void` to indicate that it is intentionally empty:

```c
void power_all(void);
```

Same below and in `.c` file.

jack-greenberg (changes requested):
Nice! Just a couple small things.

PR #160, opened 2022-07-24
Ayush/super blink tutorial

ayushchakra:


tutorials/super_blink/BUILD
```diff
@@ -0,0 +1,14 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+cc_library(
```
jack-greenberg, line 3: Should not be a `cc_library`. What should it be?

tutorials/super_blink/blink_config.h
```diff
@@ -0,0 +1,37 @@
...
+        .pin_behavior = DISCONNECTED,
+        .interrupt_enable = true,
+        .interrupt_callback = timer0_callback,
+    },
+};
+
+uint8_t can_data[] = { 0xF };
+
+can_frame_t can_msg = {
+    .id = 0,
```
jack-greenberg, line 33: While it's okay to have an ID as 0, it's not a good practice. ID 0 has the highest priority. We will go over IDs on Wednesday!

tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,36 @@
+#include "libs/timer/api.h"
+
+#include "blink_config.h"
+#include "libs/can/api.h"
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+#include <avr/interrupt.h>
+
+volatile uint8_t scaled_timer = 0;
```
jack-greenberg, line 10: `scaled_timer` doesn't actually have to be volatile! We will go over volatility on Wednesday. Also, you should make it local to the `main` function. As a rule of thumb, variables should have the smallest scope possible.

jack-greenberg (changes requested):
Couple things!

PR #161, opened 2022-07-24
[Firmware learning] Super blink tutorial

mkazc:


tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,31 @@
+#include "blink_config.h"
+#include <avr/interrupt.h>
+#include <stdbool.h>
+
+static volatile uint8_t counter = 0;
```
jack-greenberg, line 5: `counter` doesn't need to be volatile here because it's not updated in an ISR. We will cover this in more depth on Wednesday! Also, I recommend defining this variable in `main`. As a rule of thumb, give variables the smallest scope possible.

jack-greenberg (commented):
Looks good! Just one small comment.

PR #162, opened 2022-07-24
Add super blink program and comments to ADC code

amit-kumarh:
### Firmware Learning Session 2 HW

Add super blink program
- Blinks LED1 at 5hz
- Blinks LED2 at 30hz
- Sends CAN message with 'O', 'E', and 'M' at 10hz

### DIFF
--- a/tutorials/adc_bare/adc.c
+++ b/tutorials/adc_bare/adc.c
@@ -1,0 +1,66 @@
+#include "tutorials/adc_bare/api.h"
+#include <avr/interrupt.h>
+#include <avr/io.h>
+#include <stdlib.h>
+
+void (*interrupt_callback)(void) = NULL;
+
+ISR(ADC_vect) {
+    if (interrupt_callback != NULL) {
+        (*interrupt_callback)();
+    }
+}
+
+void adc_init(void) {
+    // Set the ADC Enable pin and setting the prescaler to 32 (125kHz)
+    ADCSRA = (1 << ADEN) | (1 << ADPS2) | (1 << ADPS0);
+    // Preserve the state of the Analog Reference bit and clear all other bits
+    ADCSRB &= (1 << AREFEN);
+    // Set voltage reference to AVcc with an external capacitor connected to
+    // AREF pin
+    ADMUX |= (1 << REFS0);
+}
+
+void adc_start_convert(adc_pin_e pin) {
+    // Clear pins 0 to 4 of the ADMUX
+    ADMUX &= ~0x1F;
+    // Set the pin from which to read the signal in the last 5 bits
+    ADMUX |= (pin & 0x1F);
+    // Set the ADC start conversion bit to start the conversion
+    ADCSRA |= (1 << ADSC);
+}
+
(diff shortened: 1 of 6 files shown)


tutorials/super_blink/blink_config.h
```diff
@@ -0,0 +1,31 @@
...
+    .channel_a = {
+        .output_compare_match = 0x822,
+        .pin_behavior = DISCONNECTED,
+        .interrupt_enable = true,
+        .interrupt_callback = timer0_callback,
+    },
+};
+
+// Define CAN message
+uint8_t OEM_TEXT[3] = { 79, 69, 77 };
```
jack-greenberg, line 24: Technically, if you want, you could do

```c
uint8_t OEM_TEXT[3] = { 'O', 'E', 'M' };
```

If you have a letter or number in single quotes in C, it is a `char` type (short for character). In AVR, `char`s are 8-bits, so they are the same as `uint8_t`s. I don't _think_ you'll get a compiler warning if you do, but you might...

tutorials/super_blink/blink_config.h
```diff
@@ -0,0 +1,31 @@
...
+
+gpio_t LED1 = PB0;
+gpio_t LED2 = PB1;
+
+timer_cfg_s timer0_cfg = {
+    .timer = TIMER0,
+    .timer0_mode = TIMER0_MODE_CTC,
+    .prescalar = CLKIO_DIV_64,
+    .channel_a = {
+        .output_compare_match = 0x822,
```
jack-greenberg, line 16: Timer 0 is an 8-bit timer, so it will count up to a maximum of `255`. That means when you use `0x822` as the compare match, it will actually be `0x822 & 0xFF` or `0x22`. So your frequency will be ~1838 Hz. You can either switch to timer 1 or you can use a different prescalar/output_compare_match pair.

tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,47 @@
...
+    gpio_set_mode(LED2, OUTPUT);
+
+    int timer_counter = 0;
+    while (true) {
+        if (run_blink) {
+            timer_counter++;
+            run_blink = false;
+
+            // Blink LED2 30 Hz, so every interrupt callback
+            gpio_toggle_pin(LED2);
```
jack-greenberg, line 29: `LED1`?

jack-greenberg (commented):
Looks good! A couple small things.

tutorials/adc_bare/adc.c
```diff
@@ -0,0 +1,66 @@
...
+    // Preserve the state of the Analog Reference bit and clear all other bits
+    ADCSRB &= (1 << AREFEN);
+    // Set voltage reference to AVcc with an external capacitor connected to
+    // AREF pin
+    ADMUX |= (1 << REFS0);
+}
+
+void adc_start_convert(adc_pin_e pin) {
+    // Clear pins 0 to 4 of the ADMUX
+    ADMUX &= ~0x1F;
```
amit-kumarh, line 26: now I actually understand why that throttle issue was happening

PR #165, opened 2022-07-30
Added configurable CAN mask to CAN API

ayushchakra:


projects/can_api/utils.py
```diff
@@ -1,14 +1,18 @@
 def get_rx_messages(subs, messages):
     """
-    Gets the received Messages from the list of messages, based on the subs 
+    Gets the received Messages from the list of messages, based on the subs
     dictionary.
     """
     rx_message_names = [d["name"] for d in subs]
     rx_messages = list(filter(lambda m: m.name in rx_message_names, messages))
+    masks = {
```
amit-kumarh, line 8: Suggestion: consider using `subscriber.get('mask', 0x7ff)` to replace the inline if for brevity and readability

Also, can we rename from `subscriber` to `message` or `msg`, since that's a more accurate depiction of what that object is?

projects/can_api/utils.py
```diff
@@ -1,14 +1,18 @@
...
+        subscriber["name"]: subscriber["mask"] if "mask" in subscriber.keys() else 0x7FF
+        for subscriber in subs
+    }
```
amit-kumarh, line 11: If there's a message in `subs` that isn't in our database's messages, it will still populate in the mask dictionary. The same thing happens with the mobs below. The C file will only generate based on the message names though, so the only consequence is some extra memory usage. 

@jack-greenberg  if this is something worth worrying about

amit-kumarh (changes requested):
Looks good! Few comments.

Can you update the name of the PR? Something like `Add configurable RX filters to CAN API` is a lot more descriptive than just the branch name.

projects/can_api/utils.py
ayushchakra: Ah completely agree with the first suggestion. As for the second one, isn't subscriber the appropriate name since I'm looping through subs, which is a list of potential subscribers that CAN messages can be received from?

projects/can_api/utils.py
ayushchakra: I don't think memory is a concern in this situation since it's just being used to generate a C File, and would just be thrown out if not in the messages, but I could be wrong.

richardli03 (approved):

projects/can_api/utils.py
jack-greenberg: if the message doesn’t exist, we should throw an error and compiling should fail. can you implement that?

projects/can_api/utils.py
jack-greenberg: the MCU itself is the subscriber. each entry would be a subscription. i like Amit’s idea of using message.

projects/can_api/utils.py
ayushchakra: yea, sounds good!

projects/can_api/tests/rx_mask.yml
```diff
@@ -0,0 +1,5 @@
+name: rx_only
+subscribe:
+  - name: test_tx
+    mob: 0
```
jack-greenberg, line 4: Nit: by our team's convention, MOb 0 is only used for TX. This is fine in theory, but for sake of continuity can you make it 1?

projects/can_api/files/c_templates/c_file.j2
```diff
@@ -51,11 +51,7 @@ can_frame_t {{ message.name|lower }}_msg = {
 
 can_filter_t {{ message.name|lower }}_filter = {
     .id = {{ message.frame_id }},
-
-    /*
-     * TODO: It would be nice if this was configurable...
```
jack-greenberg, line 56: Yay! i love when TODOs get erased :D

jack-greenberg (changes requested):

projects/can_api/tests/rx_mask.yml
ayushchakra: ah yes, will do

projects/can_api/utils.py
```diff
@@ -1,18 +1,24 @@
+from multiprocessing.sharedctypes import Value
```
amit-kumarh, line 1: Is this import here used for anything? If this is for the ValueError, that's a builtin, so no need to import it. This is something from the [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#shared-ctypes-objects) library which doesn't look like its being used in this project

amit-kumarh (changes requested):

projects/can_api/utils.py
ayushchakra: Ah my bad, must have accidentally imported it when trying to tab complete.

projects/can_api/tests/test_yaml_handler.py
```diff
@@ -2,38 +2,40 @@
...
 
-    assert(signals[1].name == "example_16_bit")
+    assert signals[1].name == "example_16_bit"
     example_16_bit = signals[1]
 
-    assert(example_16_bit.length == 16)
-    assert(example_16_bit.unit == "V")
-    assert(example_16_bit.offset == 0)
-    assert(example_16_bit.scale - (5 / 4096) < 0.00001) # Gotta love floats
+    assert example_16_bit.length == 16
```
jack-greenberg, line 29: Any idea why all the parentheses got removed?

projects/can_api/utils.py
```diff
@@ -1,14 +1,21 @@
...
     rx_messages = list(filter(lambda m: m.name in rx_message_names, messages))
+    masks = {message["name"]: message.get("mask", 0x7FF) for message in subs}
 
     mobs = {}
 
     for m in subs:
         mobs[m["name"]] = m["mob"]
 
-    return rx_messages, mobs
+    for mask_subscription_name in masks.keys():
```
jack-greenberg, line 15: It makes more sense to iterate through `rx_messages` I think. Same outcome, but more idiomatic.

jack-greenberg: Actually, this isn't doing anything. I don't think this will ever fail because `masks` is created from `subs` and so is `rx_message_names`, so they will always have the same items. This might not be necessary (I think the error might be caught elsewhere).

But you should try making a test for it! Create a test that tries to subscribe to a message not in the database and assert that it fails somehow.

jack-greenberg (changes requested):

projects/can_api/tests/test_yaml_handler.py
ayushchakra: The linter was flagging it as an error, not sure why.

projects/can_api/utils.py
amit-kumarh: Yeah, I think it would make sense to do this check just after line 7, and then check if each member of rx_message_names is in rx_messages (bearing in mind you'd need to to pull out the name for each message object).

I did test locally though, without some type of error handling it'll still work, it'll just not say you have an undefined message and generate masks and mobs without using them.

projects/can_api/utils.py
```diff
@@ -1,14 +1,25 @@
...
+
+
 def get_rx_messages(subs, messages):
     """
-    Gets the received Messages from the list of messages, based on the subs 
+    Gets the received Messages from the list of messages, based on the subs
     dictionary.
     """
+    # Check to see if subs has a message not specified in the DBC
+    dbc_message_names = [dbc_message.name for dbc_message in messages]
```
ayushchakra, line 10: This is the check we talked about last commit about checking for if the subs has a message that is missing from the dbc or not - not sure if it belongs here or if it should be more upstream or not though.

projects/can_api/utils.py
```diff
@@ -3,19 +3,20 @@ def get_rx_messages(subs, messages):
     Gets the received Messages from the list of messages, based on the subs
     dictionary.
     """
+    # Check to see if subs has a message not specified in the DBC
+    dbc_message_names = [dbc_message.name for dbc_message in messages]
+    for message in subs:
+        if message["name"] not in dbc_message_names:
+            raise ValueError(
+                f"Subscription message {message['name']} was not found in the dbc messages."
```
amit-kumarh, line 11: You don't have to change this at all, just a cool Python thing you might like though is this guy :=, the [walrus operator](https://realpython.com/python-walrus-operator/), basically lets you assign and return a value at the same time, so you write something like:

```
if name := message["name"] not in dbc_message_names:
    raise ValueError(f'{name} not found')
```

Anyway just a fairly new python feature that I think is cool haha.

amit-kumarh (approved):
.

amit-kumarh (commented):
Code looks good, build is failing because the air_control_updater is still chillin there, not sure if you want to remove it here or do it in another PR but either way we should remove it bc it will cause all the PRs to fail CI

amit-kumarh (approved):
LGTM!

jack-greenberg (approved):

PR #166, opened 2022-08-01
Added Fake Motor Controller Firmware

ayushchakra:


tutorials/fmc/fmc_config.h
```diff
@@ -0,0 +1,27 @@
...
+timer_cfg_s timer1_fan_cfg
+    = { .timer = TIMER1,
+        .timer1_mode = TIMER1_MODE_PHASE_CORRECT_PWM_10_BIT,
+        .prescalar = CLKIO_DIV_8,
+        .channel_a = { .output_compare_match = 80 },
+        .channel_b = {
+            .channel = CHANNEL_B,
+            .pin_behavior = SET,
+            .interrupt_enable = false,
+        } };
```
jack-greenberg, line 27: Nit: add a newline at the end of files. You might need to reconfigure your text editor. We should add something to CI to automatically check this...

tutorials/fmc/fmc_config.h
```diff
@@ -0,0 +1,27 @@
+#include "libs/adc/api.h"
+#include "libs/timer/api.h"
+
+adc_pin_e adc_pin = ADC0;
+
+void timer0_callback(void);
+
+timer_cfg_s timer0_cfg = {
```
jack-greenberg, line 8: Good practice to leave a comment about what the timer configuration is doing since it's pretty hard to tell. Talk about its purpose and how you arrived at 1249 as the number. Same for `timer1_cfg`.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+uint8_t high_voltage_counter = 0;
+uint8_t buffer_counter = 0;
+
+void timer0_callback(void) {
+    timer_50hz = true;
+}
+
+void change_duty_cycle(uint8_t torque_request_value) {
+    float duty_cycle = (float)torque_request_value / 255.0;
+    OCR1B = (uint16_t)duty_cycle * 80.0;
```
jack-greenberg, line 21: Where did 80 come from? Anytime you have numbers appear in your code, you should move them to `#define` statements and leave a comment wherever you define them about their purpose. Just having `80.0` in there is called a "magic number" because it is magic and it's hard to tell what it means.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
+#include "fmc_config.h"
+#include "libs/adc/api.h"
+#include "libs/timer/api.h"
+#include "tutorials/fmc/can_api.h"
+
+#include <avr/interrupt.h>
+#include <avr/io.h>
+
+volatile bool timer_50hz = true;
+uint16_t adc_value = 0;
```
jack-greenberg, line 10: Why is this global?

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+bool high_voltage = false;
+uint8_t high_voltage_counter = 0;
+uint8_t buffer_counter = 0;
```
jack-greenberg, line 13: Why are any of these global? Seems like they should be local to main. Remember: always give variables the smallest possible scope.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+}
+
+void change_duty_cycle(uint8_t torque_request_value) {
+    float duty_cycle = (float)torque_request_value / 255.0;
+    OCR1B = (uint16_t)duty_cycle * 80.0;
+}
+
+int main(void) {
+    sei();
+    timer_init(&timer0_cfg);
```
jack-greenberg, line 26: Where are you initializing timer1?

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+                change_duty_cycle(0);
+                buffer_counter++;
+                if (buffer_counter > 49) {
+                    high_voltage_counter = 0;
+                }
+            } else if (adc_value > 2.5) {
+                change_duty_cycle(0);
+                if (buffer_counter == 0) {
+                    high_voltage_counter++;
+                }
```
jack-greenberg, line 49: I'm confused about why you have a buffer counter and also a high voltage counter. Also, high voltage counter doesn't make sense as a variable name I don't think because it's not monitoring high voltage anything. The ADC is supposed to be hooked up to a temperature sensor.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+    sei();
+    timer_init(&timer0_cfg);
+    can_init_fmc();
+    adc_init();
+
+    can_receive_motor_command();
+
+    while (1) {
+        if (timer_50hz) {
+            adc_value = adc_read(adc_pin);
```
jack-greenberg, line 34: `adc_pin` isn't a very clear variable name. I usually name ADC pin variables the same way I'd name a GPIO variable (like `TEMPERATURE_SENSE` or something).

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+            fmc_data[0] = adc_value & 0xFF;
+            fmc_data[1] = (adc_value >> 8);
```
jack-greenberg, line 36: Because we are using the CAN API, you can actually just do `fmc.signal_name = variable_name`.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+            can_send_fmc();
+
+            if (high_voltage_counter > 49) {
+                change_duty_cycle(0);
+                buffer_counter++;
+                if (buffer_counter > 49) {
+                    high_voltage_counter = 0;
```
jack-greenberg, line 43: More magic numbers.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+                }
+            } else if (adc_value > 2.5) {
+                change_duty_cycle(0);
+                if (buffer_counter == 0) {
+                    high_voltage_counter++;
+                }
+            } else if (can_poll_receive_motor_command() == 0) {
```
jack-greenberg, line 50: Why is this in an `else if`? That means this won't execute if either of the two branches are true.

tutorials/fmc/fmc.c
```diff
@@ -0,0 +1,59 @@
...
+                change_duty_cycle(motor_command.torque_request);
+                can_receive_motor_command();
+                high_voltage_counter = 0;
+                buffer_counter = 0;
```
jack-greenberg, line 54: Also why does receiving a motor command reset these two counters?

tutorials/fmc/fmc.yml
```diff
@@ -0,0 +1,16 @@
+name: fmc
+subscribe:
+- name: motor_command
+  mob: 1
+publish:
+- name: fmc
+  id: 0xC
+  freq_hz: 50
+  signals:
+    - name: adc_value
```
jack-greenberg, line 10: I believe this was supposed to be a temperature sensor, so I recommend giving it a more descriptive name.

tutorials/fmc/fmc_config.h
```diff
@@ -0,0 +1,27 @@
+#include "libs/adc/api.h"
+#include "libs/timer/api.h"
+
+adc_pin_e adc_pin = ADC0;
```
jack-greenberg, line 4: Give this a more descriptive name.

jack-greenberg (changes requested):

tutorials/fmc/fmc_config.h
jack-greenberg: Also, timer1 is called `timer1_fan_cfg` but it doesn't control a fan...

PR #167, opened 2022-08-09
HitL Software Revamp

amit-kumarh:
This PR makes some updates to the HitL software stack in preparation for full HitL integration. Major changes: 

- Removed a lot of the static configuration files in favor of passing in Pytest fixtures or (in the case of the DBC) retrieving through Bazel
- Refactor IOController into two classes:
  - The IOController will be the user facing class, which will take in a pin_info dictionary and provide access to the set_state and get_state methods.
  - That class will then call on FT422Proxy class, which will interface with the actual I2C hardware.
- Other small adjustments and minor cleanup

#### Revision 2
- Improve CAN Controller docstrings
- Integrate unit tests into Bazel
- Improve CAN Controller error handling
- Bug fixes/cleanup

#### Revision 3
- Refactor to remove HitL class and we will directly access the CAN and IO controllers
- Added fatal errors when hardware isn't connected since we're now accessing each class separately
- Updated unit test conftest to reflect this (see for examples for use in HitL tests
- BUILD file changes


amit-kumarh:
Not ready for review just yet but looks like there's a CI bug... if we delete a python file, black tries to scan it because it's a changed file but because we deleted it it no longer exists

projects/hitl/software/hitl/ft4222_proxy.py
```diff
@@ -0,0 +1,199 @@
...
+    0: 0b00000000,
+    1: 0b00001000,
+    2: 0b00000001,
+    3: 0b00001001,
+}
+ADC_PREFIX = 0b10110000
+ADC_RETURN_SIZE_BYTES = 3
+
+# DAC parameters
+DAC_CHANNEL_TO_ADD_BITS = {i: i for i in range(8)}
```
amit-kumarh, line 29: Is this and line 37 necessary? It was in the original code, but is there any reason we use the pin number directly in lines 84, 115, and 190?

projects/hitl/software/hitl/iocontroller.py
```diff
@@ -1,302 +1,102 @@
...
-
-            # Wait for response
-            response = self.dev.i2cMaster_Read(address, ADC_RETURN_SIZE_BYTES)
-            self.log.debug(f"Received {response}")
-            out = self._map_to_human(response, self.pin_info[name]["min"], self.pin_info[name]["max"])
+        if self.pin_info[name]["type"] == PinType.ANALOG:
+            min = self.pin_info[name]["min"]
+            max = self.pin_info[name]["max"]
+            out = self.ft4222.get_analog(address, pin_number, min, max)
         else:
```
jack-greenberg, line 94: I would do `elif type == PinType.DIGITAL` and then have an `else` that raises a fault with invalid type.

projects/hitl/software/hitl/utils.py
```diff
@@ -4,48 +4,123 @@
...
+        "address": address,
+        "pin": num,
+        "type": type,
+        "read_write": visibility,
+        "min": min,
+        "max": max,
+    }
 
 
 def find_arduino() -> Tuple[Optional[str], Optional[str]]:
```
jack-greenberg, line 126: Deprecated, so remove?

projects/hitl/software/tests/mkv.dbc
```diff
@@ -0,0 +1,495 @@
+VERSION ""
```
jack-greenberg, line 1: This file shouldn't be tracked in Git. It should be loaded in as a `data` dependency.

projects/hitl/tests/conftest.py
```diff
@@ -1,7 +1,18 @@
 import pytest
```
jack-greenberg, line 1: Remove this folder (`projects/hitl/tests`) for now? I thought we decided to put tests in a different folder?

jack-greenberg (changes requested):
Awesome stuff! Left a few comments, and I'll re-review after they're addressed.

projects/hitl/software/hitl/cancontroller.py
```diff
@@ -16,200 +11,190 @@
...
+                },
             )
-            listener.daemon = True
             listener.start()
-        else:
-            can_bus = None
-            self.log.error("Not on linux; initializing self.can to None")
+        except OSError as e:
+            self.log.error("CAN Hardware not detected - initializing to none")
+            self.can_bus = None
```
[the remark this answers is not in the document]
amit-kumarh: So this was in the old code, the rationale here is that you can still run a HitL object even if you don't have both the IO device and the CAN device connected if your tests only need one or you're just testing a subset of functionality or something like that.

So it gets initialized to None and then in all of the functions there's a quick check to see if it's None and if it is than it just logs an error and returns but doesn't crash the program. There's a similar thing in the FT4222 proxy as well

projects/hitl/software/hitl/cancontroller.py
```diff
@@ -16,200 +11,190 @@
...
+        Args:
+            path (str): Path to CAN spec file (DBC)
         """
         # Create database that has all the messages in the dbc file in type message
         self.db = cantools.database.load_file(path)
 
         # Iterates through messages to create state dictionaries
         message_of_signal = {}
-        signals = {}
+        signals = defaultdict(lambda: dict())
```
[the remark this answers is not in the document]
amit-kumarh: Defaultdict is a type in Python that takes a callable, and then if you try to access a key in it that doesn't exist, it creates with a default value of whatever the callable returns. So in line 189, when we run `signals[msg.name][sig.name] = 0`, with a regular dict we'd get a KeyError because `signals[msg.name]` doesn't exist, but the defaultdict will see that and automatically initialize `signals[msg.name]` to an empty dict.

However, I just looked it up and turns out it doesn't actually have to be a lambda, since dict is a callable on its own, the lambda is just for when you'd need to pass some parameters in

amit-kumarh:
Rev 2 with those comments addressed. Unit tests worked with Bazel but only when I changed tools/pytest/requirements.txt to let me use a pytest version higher than 6.1.2 - there are some issues there with newer versions of Python that didn't get fixed until later pytest versions. Felt it was outside the scope of this PR but something that you should be aware of. Also, they are currently set to fail out after one test fails and not run any more, not sure if you wanted it to behave that way.

projects/hitl/software/tests/conftest.py
```diff
@@ -1,8 +1,35 @@
...
+def hitl(bspd, dbc):
+    hitl = Hitl(bspd, dbc)
+    yield hitl
+
+    hitl.close()
+
+
+@pytest.fixture
+def dbc():
+    return "tests/mkv.dbc"
```
jack-greenberg, line 18: Is this really the correct path? Shouldn't it be `vehicle/mkv/mkv.dbc`?

amit-kumarh: Yep, thanks. Changed it in the CAN unit tests but didn't make it call on the DBC fixture so I missed changing this one.

projects/hitl/software/hitl/ft4222_proxy.py
```diff
@@ -0,0 +1,199 @@
...
+import os
+
+from utils import map_to_human, map_to_machine
+
+try:
+    import ft4222
+except ModuleNotFoundError:
+    # We haven't setup the ft4222 library yet...
+    # See https://awenstrup.github.io/setup.html
+    ft4222 = None
```
ayushchakra, line 15: Can anything be done if the FT4222 library isn't set-up? If not, we might want a more fatal error that terminates execution.

projects/hitl/software/hitl/ft4222_proxy.py
```diff
@@ -0,0 +1,199 @@
...
+
+# GPIO parameters
+GPIO_CHANNEL_TO_ADD_BITS = {i: i for i in range(4, 32)}
+GPIO_COMMANDS = {"single port": 0b00100000}
+GPIO_RETURN_SIZE_BYTES = 1
+
+
+class FT4222Proxy:
+    """
+    Proxy class to intergace with FT4222 chip
```
ayushchakra, line 44: interface*

ayushchakra (changes requested):
I just made two small comments, but other than that, it looks good to me!

projects/hitl/software/hitl/ft4222_proxy.py
amit-kumarh: So this was in the old code, the rationale here is that you can still run a HitL object even if you don't have both the IO device and the CAN device connected if your tests only need one or you're just testing a subset of functionality or something like that.

So it gets initialized to None and then in all of the functions there's a quick check to see if it's None and if it is than it just logs an error and returns but doesn't crash the program. There's a similar thing in the CAN Controller as well. Open to changing it though, that's not something that's set in stone

jack-greenberg (approved):
Looks good! I guess there will be changes once we get hardware but i'm fine to merge this. feel free to ignore the CI issue regarding the deleted python file. we should figure that out another time...

ayushchakra (approved):
Looks good to me, nice work!

PR #170, opened 2022-08-28
Add enum generation to CAN API

jack-greenberg:
Wrote this code a while ago but forgot about it. Not sure if it still has merit, but figured I'd post for posterity's sake.

---

This PR adds generation of C enums to the CAN API. Previously, we had duplicated information: we defined enums in both the YAML file of the CAN API *and* in the C source code, meaning that if one was updated, the other had to be updated as well. Now, we will be able to just use the YAML as a single source of truth and the generated code has enums that the C source can use. I will post an example here in a few days, but in the meantime I updated all the firmware images to implement this new feature.

### DIFF
--- a/projects/can_api/dbc_generator.py
+++ b/projects/can_api/dbc_generator.py
@@ -5,32 +5,31 @@ from yaml_handler import YamlParser
 
 __version__ = "2021.12.28"
 
+
 def main():
     """
     CLI to generate a DBC file from YAML files
     """
     parser = argparse.ArgumentParser(
-        description = "Generate DBC from YAML and DBC files",
+        description="Generate DBC from YAML and DBC files",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
 
     parser.add_argument(
-        '-o', '--output',
-        default='out.dbc',
-        help='Output filename (default is out.dbc)'
+        "-o", "--output", default="out.dbc", help="Output filename (default is out.dbc)"
     )
 
     parser.add_argument(
-        'inputs',
-        help='Input YAML and DBC files',
+        "inputs",
+        help="Input YAML and DBC files",
         nargs="+",
     )
 
     args = parser.parse_args()
 
     db = CANDatabase(
-        strict = True,
-        version = __version__,
+        strict=True,
+        version=__version__,
     )
 
     # Add in extra DBC files
@@ -45,5 +44,6 @@ def main():
 
     dump_file(db, args.output)
 
+
 if __name__ == "__main__":
-  main()
+    main()
--- a/vehicle/mkv/software/air_control/air.c
+++ b/vehicle/mkv/software/air_control/air.c
@@ -407,7 +407,7 @@ int main(void) {
             run_1ms = false;
         }
 
-        if (air_control_critical.air_state == IDLE) {
+        if (air_control_critical.air_state == AIR_STATE_IDLE) {
             updater_loop();
         }
 
--- a/vehicle/mkv/software/air_control/air.c
+++ b/vehicle/mkv/software/air_control/air.c
@@ -233,7 +233,7 @@ static void state_machine_run(void) {
 
                 if (motor_controller_voltage
                     > (PRECHARGE_THRESHOLD * pack_voltage)) {
-                    gpio_set_pin(AIR_N_LSD); // Close AIR_N
+                    gpio_set_pin(AIR_N_LSD); // Close AIR negative
                     gpio_clear_pin(PRECHARGE_CTL); // Close precharge relay
                     once = true;
                     air_control_critical.air_state = AIR_STATE_TS_ACTIVE;
@@ -261,7 +261,7 @@ static void state_machine_run(void) {
 
             /*
              * This pattern ensures that we only call get_time() once because we
-             * only want to capture the time that PRECHARGE starts
+             * only want to capture the time that DISCHARGE starts
              */
             static bool once = true;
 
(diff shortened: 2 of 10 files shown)


vehicle/mkv/software/air_control/air.c
```diff
@@ -262,7 +236,7 @@ static void state_machine_run(void) {
                     gpio_set_pin(AIR_N_LSD); // Close AIR_N
```
bubbleebubbles, line 236: nit: expand comment to say AIR Negative instead of just AIR_N

vehicle/mkv/software/air_control/air.c
```diff
@@ -274,15 +248,15 @@ static void state_machine_run(void) {
...
             if (!air_control_critical.ss_tsms) {
-                air_control_critical.air_state = DISCHARGE;
+                air_control_critical.air_state = AIR_STATE_DISCHARGE;
             }
         } break;
-        case DISCHARGE: {
+        case AIR_STATE_DISCHARGE: {
             gpio_clear_pin(AIR_N_LSD);
 
             /*
```
bubbleebubbles, line 262: on line 264, do you mean DISCHARGE?

vehicle/mkv/software/air_control/air.c
```diff
@@ -341,22 +315,22 @@ static void state_machine_run(void) {
...
             }
         } break;
-        case FAULT: {
+        case AIR_STATE_FAULT: {
             gpio_set_pin(FAULT_LED);
             gpio_clear_pin(PRECHARGE_CTL);
             gpio_clear_pin(AIR_N_LSD);
         } break;
         default: {
             // Shouldn't happen, but just in case
```
bubbleebubbles, line 332: can you expand on what you mean by "just in case". I know you mean that we want to make sure we don't accidentally start precharge when something is wrong but I think it would be nice to include :)

vehicle/mkv/software/bms/utils/fault.h
```diff
@@ -1,51 +1,52 @@
...
-    BMS_FAULT_PEC,
-};
-
-void set_fault(enum FaultCode the_fault);
+#include "vehicle/mkv/software/bms/can_api.h"
+
+// enum FaultCode {
+//     /*
+//      * Normal operation
+//      */
```
bubbleebubbles, line 7: I really like the comments about what each fault means. :)

projects/can_api/dbc_generator.py
```diff
@@ -5,32 +5,31 @@
...
-        description = "Generate DBC from YAML and DBC files",
+        description="Generate DBC from YAML and DBC files",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
 
     parser.add_argument(
-        '-o', '--output',
-        default='out.dbc',
-        help='Output filename (default is out.dbc)'
+        "-o", "--output", default="out.dbc", help="Output filename (default is out.dbc)"
```
bubbleebubbles, line 19: most places you expand arguments onto separate lines while here you're condensing it onto one. Might be better to stay more consistent and do one or the other.

vehicle/mkv/software/dashboard/dashboard.c
```diff
@@ -112,8 +112,7 @@ int main(void) {
 
         // BMS Core message for BMS Status LED
         if (can_poll_receive_bms_core() == 0) {
-            if (bms_core.bms_fault_state
-                != BMS_FAULT_NONE) { // check BMS status
+            if (bms_core.bms_fault != BMS_FAULT_NONE) { // check BMS status
                 gpio_set_pin(BMS_LED); // set BMS light high
             } else {
                 gpio_clear_pin(BMS_LED); // BMS OFF means NOT OK
```
bubbleebubbles, line 118: BMS LED should be on if things are not ok, i think the comments need to flip.

projects/can_api/dbc_generator.py
bubbleebubbles, line 19: most places you expand arguments onto separate lines while here you're condensing it onto one. Might be better to stay more consistent and do one or the other.

vehicle/mkv/software/air_control/air.c
bubbleebubbles, line 236: nit: expand comment to say AIR Negative instead of just AIR_N

vehicle/mkv/software/air_control/air.c
bubbleebubbles, line 262: on line 264, do you mean DISCHARGE?

vehicle/mkv/software/air_control/air.c
bubbleebubbles, line 332: can you expand on what you mean by "just in case". I know you mean that we want to make sure we don't accidentally start precharge when something is wrong but I think it would be nice to include :)

vehicle/mkv/software/bms/utils/fault.h
bubbleebubbles, line 7: I really like the comments about what each fault means. :)

bubbleebubbles (commented):
Everything looks much cleaner and easier to read now!

projects/can_api/dbc_generator.py
jack-greenberg: Annoyingly, this is the linter's doing. The linter will give an error if I change it lol

vehicle/mkv/software/air_control/air.c
jack-greenberg: yep!

vehicle/mkv/software/air_control/air.c
jack-greenberg: What i mean't by "just in case" is that the case statements for the enum should be exhaustive and catch every case that `air_state` could be, but in case there is some weird memory issue, we have a default for the switch statement that sends us into fault.

amit-kumarh (approved):
LGTM! Thanks for working on this!

PR #171, opened 2022-09-11
Add charging support to BMS

jack-greenberg:


jack-greenberg:
Closing this because I think we should just revisit charging support from scratch in the spring.

PR #174, opened 2022-09-19
example schematic for onboarding

bubbleebubbles:
sample LED with push button schematic for onboarding to teach first years how to use kicad.


jack-greenberg:
Since I don't think we're using this, gonna close it. But if you want it approved and merged, just let me know and we can do it!

PR #175, opened 2022-09-30
Add M.2 Footprint and Board Examples

jack-greenberg:
# Board connector footprint

image

# Edge connector footprint

image


jack-greenberg:
Daughter board looks great. I'd recommend adding test points for power and ground so that if there's an issue with the motherboard, we can still debug. Maybe at least power (5V) and ground. Also, do you want to just add a row of LEDs so that we can test all of the LED colors at once in a row? Like connect all of the PBx pins to a different colored LED?

jack-greenberg:
image

Nit: I would switch this so SCK doesn't go through a via since it is higher frequency than the CAN_TERM_EN signal.

jack-greenberg:
For the mounting hole in the motherboard, should we spec out a component with the correct size hole and threads for a standard screw?

Batt+ trace should be thicker if possible. I usually use 30 mils I think.

I would switch this so that 5V doesn't go through a via and PD7 does.

image

elviswolcott:
https://www.digikey.com/en/products/detail/keystone-electronics/4939/9991052
thinking of using these solder in threaded inserts. once we have the board in hand confirm clearances we can go ahead and order 4/40 standoffs and screws from mcmaster to completed the setup

bubbleebubbles (approved):

PR #178, opened 2022-11-03
Throttle Schematic First Draft

eys123:


vehicle/mkvi/hardware/nostril/throttle/throttle/BUILD
```diff
@@ -0,0 +1,5 @@
+load("//bazel:defs.bzl", "kicad_hardware")
+
+kicad_hardware(
+    name = "template_board",
```
jack-greenberg, line 4: Be sure to update this BUILD file so that it matches your schematic!

jack-greenberg (dismissed):
Looks good Ellen! A couple things
* You have two nested `throttle` folders. This usually happens because KiCad will by default create a new folder for your project. Just make sure to move your KiCad files up one folder. Happy to show you what that means in person.
* This schematic looks super clean! Great job
* Currently, your CAN+ and CAN- don't leave the Microcontroller sheet. They'll need to be hierarchical labels and then connected to the M.2 connector
  * Also, on your connector, they're called CANH and CANL, but they should be `CAN_+` and CAN_-`.
* It seems like the shutdown circuit ends at your schematic (you have a SHDN_IN which you call SS_InertialSwitch, but no SHDN_OUT). You'll need to make some modifications to your Shutdown Sense block to make sure we have a complete shutdown circuit! :) 
* Your "Throttle Potentiometer" block is a little confusing. It's just a weird thing to capture since you have off-board components. I might recommend not having a block for those and just putting them on your top=level schematic and maybe don't include the potentiometer symbol (maybe put pictures (PNGs) of potentiometers instead? IDK)
* How come you're using global labels for the potentiometer power and ground instead of hierarchical labels?
* Can you switch the potentiometer inputs to PB2 and PB5 on the microcontroller? It will make things easier to test with HitL since I think Hitl can't test ADC9 and ADC10 (PC4 and PC5) at the same time :/

Other than that, make sure to run the ERC (electrical rule checker) and make sure you don't have any serious issues. Ask Amit about this :)

Great job!

elviswolcott (approved):
Looks like all of the requested changes were made. This is a really clean schematic - great work so far!

PR #179, opened 2022-11-04
J1772 Charging Support

meaganlipsman:
First iteration of the schematic for the J1772 charger. This board is trying to facilitate communication between EVSE(outside charger) and the accumulator. Also manages CAN messages between the vehicle and the charger.


elviswolcott:
I added some feedback after going through the checklist to your Coda page. This is looking great and I'm super excited for how this project is turning out! I think we should be able to go through everything at the VIO meeting today and (maybe) get you started on PCB design!

elviswolcott (approved):

PR #180, opened 2022-11-06
Add IMD latch board layout

MPietraski:


jack-greenberg (changes requested):
This looks great! I like how organized the schematic is. It's clear what the different blocks are. It's a bit unclear what is connected to what, but it's tough to balance that with messiness, so I think it's fine.

## Notes

* I would like to see calculations for time delay (why are you using those resistor values? What is the delay? Etc)
  * Same for **Read IMD**
* Why are you using two different low-side drivers? Why not just use one with two different loads?
* For the LED, can you include text about what the LED means? Like you can say "If lit, IMD is OK. If off, IMD has tripped"

Nice job Miranda!!!

bubbleebubbles:
Great work overall Miranda! The schematic looks awesome! I have a couple additional notes on top of jack's comments. Let me know if you have any questions :) 

## Feedback:
* Would love to see a note about the expected voltage at the inputs of the op amps after each voltage divider (or voltage range). This helps when testing if you need to refer to schematic quickly. 
* U3 (CD4044BDR) is missing the datasheet, purchasing link, footprint, MPN, MFN. 
* For the S/R latch, it would be useful to include the truth table on the schematic based on the IMD output and time delay. In addition to that, a screenshot from the datasheet of the expected truth table would also be helpful. 
* Sort of similar to jack’s comment -- can you add a note below the IMD status section saying what the IMD outputs when healthy/not healthy. (IMD outputs 12V when it’s healthy) 
* I see that the part shows an X for a DNC (Do not connect) but i’m not sure Kicad recognizes that. Make sure to add the kicad symbol for DNC as well, see the image below. 
DNC
* The TLV3701--- comparator ((U1 and U2) is out of stock :( 
comparator

I really love how clean the schematic looks, it's far easier to read now than it was in the past, haha :D Awesome job!

jack-greenberg:
Looks great! Somehow you pulled in the HitL changes, so you'll need to get rid of those. Melissa or Elvis can help with that. Also, can you update your filenames to follow our naming convention of all lower-case with underscores between words? So instead of `IMD_Latch.kicad_pro` it will be `imd_latch.kicad_pro`. Same thing goes for your folder name (looks like you used a dash instead of an underscore). 

Great job Miranda!

jack-greenberg (changes requested):
Just a couple small things :)

bubbleebubbles (approved):
Everything looks great!

bubbleebubbles (approved):

jack-greenberg (changes requested):
This is super close! I just have a few comments:

image
* Your ground plane should cover the entire board--no need to cut it off here. This will also let you put your ground connections in the M.2 board closer to the edge connector itself.
image
* The edge of your board should bisect the mounting hole
image
* You use two power pins because the shutdown circuit has slightly high current, but you immediately break it out into just a single 6mil trace. The shutdown circuit must carry enough current to close the AIRs. That is about 3.8A instantaneously. A 6mil trace will likely get damaged after enough cycles with the AIRs, so I would recommend making your SHDN_IN and SHDN_OUT 30mils.
* You BATT+ and +5V supply traces should all be at least 10mils, preferrably 15mils.
* I'd recommend adding some silkscreen to your board to indicate the name of the board, the author, the revision (Rev 1), the date (just today's date is fine), and the vehicle (MKVI)
image
* Out of curiosity, why do you have vias here? You have a few signals with random vias that aren't needed
---
Again, awesome job! This is super super close, just needs a few tweaks.

MPietraski:
Hi, thanks for the feedback! I think I've implemented the changes requested.
The vias you pointed out are there because KiCad seems to want me to connect both the pads on the front and back copper of the M.2 connector, but I can remove them if they're not actually necessary.

bubbleebubbles:
Looks great so far! One thing to add, could you label each of your test points with the signal that they correspond to. It'll make your life much easier when debugging if you don't have to keep looking back and forth between the schematic and the board.

jack-greenberg (approved):
Looks good!

PR #181, opened 2022-11-07
Add Dashboard Schematic

amandalchang:
Created entire schematic, which includes three LSDs in hierarchical sheets, a 5V buck converter, and my estop shutdown sense.


vehicle/mkvi/hardware/nostril/dashboard/BUILD
```diff
@@ -0,0 +1,5 @@
+load("//bazel:defs.bzl", "kicad_hardware")
+
+kicad_hardware(
+    name = "template_board",
```
jack-greenberg, line 4: Can you update the BUILD file to match your project (ask Amit for examples) so that this passes CI properly :)

jack-greenberg (changes requested):
Awesome job! This schematic looks super clean and is really easy to read. Couple things:

image
* We need to do this now :/ I can help or you can talk to Amit or Dasha or Elvis
* Your CAN+ and CAN- aren't connected to anything. 
* Make sure to run ERC (electrical rule checker) and try and resolve as many issues as you can. Ask your subteam lead for help if you are unclear about this.

Other than that, looks great! Awesome job :)

jack-greenberg:
Is this supposed to have a buck converter on it? I thought we discussed removing it?

jack-greenberg:
This layout looks great! Love the package shape. Couple notes:

# Schematic
* You should have a pull-down resistor on `RTD_BUTTON_SENSE-` so that it isn't floating when the button is not pressed
* R15 should be R_120_DNP
* In the MCP25625 schematic sheet, MISO should be an output hierarchical label, its currently an input (make sure to update the pins on the top-level schematic too)
* Do we know how much current the buzzer and dashboard LEDs draw? If it's anywhere near 500mA, we should use multiple pins on the M.2 for the low-side connections. Might also want to make the traces a bit wider, maybe 10mils instead of 6.
* The *_SHDN_SENSE pins should be outputs (make sure to update the pins on the top-level schematic too)
image

# Layout
* Most of this looks really good.
* In general, you've got a lot of room on this board, so I'd recommend spacing things out a little more so that 
* Image below, you should give each pad on the crystal its own GND via for signal integrity. Also, the load capacitors and 3V3 bypass capacitor should also have their own GND vias too
image
* Same here (below). In general, use one via per ground pad
image

jack-greenberg (changes requested):
See prev comment

amit-kumarh:
Buzzer and panel lights are both <80 mA so should be fine on that front

jack-greenberg:
Leaving this here for posterity's sake: @amit-kumarh and I chatted on Slack yesterday after realizing that the current footprint and placement for the Pi Zero means that it will be sticking off of the dashboard PCB instead of hovering right over it.

It will work, and we're definitely not pressed for space, so we can leave it be, but we can also re-place and re-route the board so that it's a bit cleaner. Either way, Amit will switch gears to focus on the motherboard and I might reroute it if I get the time, otherwise we can just merge. If anyone else has thoughts put them here!

amit-kumarh:
Oh! One more thing - trace angle got a little wonky here, mind just making those 45 degree angles?

trace_angle

jack-greenberg:
* For the crystal, I kept the distance I did because I didn't want the CAN signals to interfere with them. From other crystal designs I've seen, this shouldn't cause any issues
* Ah yeah, for the vias, I found that smaller vias worked better for the design, but I'll double check the Coda page
* W/r/t the ground test pad, yeah I struggled a bit with that. The issue is that having all of the test points close together made routing super difficult, so I decided to prioritize routing

jack-greenberg:
w/r/t the via sizing and trace clearance, I think it should all be within the capabilities of Calpak, although they don't publicly list their capabilities (nowhere that I can find lol). I'll update the minimums on KiCad to get rid of the warnings if that's okay?

elviswolcott (approved):

PR #182, opened 2022-11-08
Rev 1 Precharge Discharge schematic

redhotchili0909:


jack-greenberg:
Looks good! Just a few small remaining things:

image
The schematic you have for the discharge relay looks a bit confusing, as in it's hard to see what is driving and where GND is, etc. You should use the GND symbol (the upside-down triangle). Something like this might be clearer:
image

Similar for the precharge relay. Other than that, make sure to run the KiCad Git filters (ask Melissa about that) and make sure your build passes CI (also ask Melissa), and then we can merge!

jack-greenberg (dismissed):
See comments above ^ Nice work!

elviswolcott (approved):
As of the latest changes, looks ready to merge! Clean, easy to read schematic. @mkazc can continue working with you on layout at the next HV meeting.

PR #183, opened 2022-11-08
added service section controller rev 3 schematic

Izziabby:
first commit
changed hv buck output to 5v
removed hv lin reg
removed rj45 programming header
replaced isoSPI cutout connector to micromatch


Izziabby:
Some questions/roadblocks I have for my layout:
1. What is +5V_TSAL connected to? It looks like its own unique label.
2. Should there be an independent labeling system for LV battery 5V vs accum 5V? KiCAD wants to connect them...
3. Why is TP10 bigger than the other test points next to it? 
4. Can you create zones without a 'proper' net label only the one KiCAD assigns to it (eg. Net - (COUT1-Pad1)?

PR #184, opened 2022-11-08
TSAL/AIL Layout

VintageBikes:
first draft, change connector to M.2, changed hv output to 5v, removed hv linreg


bubbleebubbles (dismissed):
Great job so far! Please see some feedback below: 
## TSAL/AIL Layout Feedback
* This board will be stepping down from 400V to 5V, which is a lot. We want to make sure that we're dissipating a lot of the heat and that we're getting good performance out of this circuit. On the [buck converter datasheet page 16](https://www.monolithicpower.com/en/documentview/productdocument/index/version/2/document_type/Datasheet/lang/en/sku/MP9488/document_id/3748/) they have some layout guidelines that I strongly recommend to follow. Try your best to arrange the components and add the copper pours they suggest so that we can get good porformance out of this power regulator. 


* HV and LV need to be separated at lest 12.7mm according to the rules and clearly marked. Make sure to add a silkscreen with the distance between the two. I don't think it's separated enough at the moment (your copper pours on the back are 10-ish mm apart). 



* Bypass capacitors need to be placed as close as possible to the power supply pin of the part they're meant for and connected without any interuptions. 


* Make sure to label all your test points clearly with the signal. You did that for some of them but not all. 

* These resistors (R1, 4, 7, 8) have gotten so hot that they've melted off in the past. These need to be spread out and need to add a copper pour underneath to help with heat dissipation. 


Great first pass! Feel free to shoot me a slack message if any of this is confusing. Thanks for the hard work!

jack-greenberg (changes requested):
Looks like there are some DRC errors to do with the board outline :/

bubbleebubbles:
> Looks like there are some DRC errors to do with the board outline :/

yee, working on that lol :D

elviswolcott (approved):

PR #185, opened 2022-11-08
MKVI BMS Schematics

mkazc:
There's an issue with `github-actions` - ignore generated files


bubbleebubbles (changes requested):
Please see below for my feedback for the Peripheral Board :) Overall, amazing job! Everything is very well labeled and well organized which made it easy to understand. 
## Peripheral Feedback
* Where on the schematic do you connect C0 to GND so that the bms chip can read this? Is this supposed to say C0 instead of GND?
cell0
* What’s the reason for an led on each cell to indicate whether it is balancing or not? It doesn’t hurt but considering there would be 18 leds + current limiting resistors per board, it might take up a significant amount of space. It could be helpful for debugging but we can also read the cell voltages using CAN which would show us which cells are balancing. 
* Follow-up: have you checked what the ESR is for the discharge resistor and the LED+current limiting resistor. It may be useful to see how that alters the balance current. 
* Make sure to DNP RX6 so that when we send the boards to the manufacturer, they know not to populate them. I’m refering to the 0 ohm jumper between CN and CN-1 
* Your board is powered using the topmost cell (C18). Make sure to add a note somewhere visible on the schematic that says that C18 always needs to be connected to the topmost cell of the pack. 
* I think it is very smart to add a 200mA fuse to protect the BMS chip FETs :) 

It's lookin good :D I'm excited for the layout!

bubbleebubbles (approved):
BMS Core: Overall, great job! Similar to the Peripheral, I think you did a great job with labeling the schematic and providing the relevant calcs. :) Please see below for some feedback, but these are mostly nitpicky comments. 
## BMS Core Feedback
* Can you label what VREF is expected to be on this part of the schematic, or what the range of values corresponds to (what current we expect to see). 
vref

* *nit Move BSPD_CURRENT_SENSE label to the left to avoid it looking like its supposed to be a part of the communication block 
nit1_pin

bubbleebubbles (changes requested):
Everything else looks great! Thanks for the hard work. Requesting changes for the previous comments :)

PR #189, opened 2022-11-11
Add layout for AIR control

vrhinehart:
Schematic should be ready for layout


jack-greenberg:
This looks good! I have a few notes that I think will help with organization and readability:
* I don't think the CAN transceiver should be in its own schematic. You could maybe have a schematic that has both the CAN transceiver and the microcontroller, but I don't think they need to be separated
  * On a similar note, the CANH and CANL labels should be CAN_+ and CAN_- (this is so that KiCad knows they are a _differential pair_ and can be routed as such, and don't need to be global labels (they should either be hierarchical or local, depending on where the can transceiver is)
* You have symbols for the AIRs on your board, which is fine, but they need to be marked as not on your layout, because the AIR's are off-board components. And furthermore, how are you connecting to the AIRs? Is it through the M.2? In that case, you'll need a lot of contacts because they are high current connections.
* For shutdown sensing, I might recommend just having a single hierarchical schematic with the 5 high-side driver circuits in there as it might be more readable
* For the hierarchical labels, "Out to micro" and "In from relay" shouldn't be used as the labels because they are all the same. I would recommend something like `IMD_SHUTDOWN_SENSE` and `IMD_SHUTDOWN`, respectively (see image below)
image
* Your programming LEDs are reversed, and should be rotated
* For any net label, please use the naming convention of all caps and underscores (i.e. AIR_P_WELD_DETECT instead of "Air + Weld Detect"
* For your 5V, you have a global label, but there is a symbol that looks like an arrow pointing up that you should be using for 5V
* For your question about shutdown being connected through the M.2, that is a question for @elviswolcott and depends on where the connections are on the motherboard. Basically, the motherboard might take care of the shutdown connections and just pass the branches used for sensing to your board.
* For the connections to your M.2, can you add net labels so it's clear what those connections are for. Right now it is a bit ambiguous
image
* It looks like a lot of your components have deleted references (i.e. R2, C102). These are needed by KiCad and help identify what components are where. You might need to delete and re-add these components. And don't forget to annotate your schematic!
image

Nice work Vaughn! Happy to re-review once you make these changes.

jack-greenberg (changes requested):
See previous comment! :)

vrhinehart:
@jack-greenberg 

It looks like you're looking at an old version. I did not update any of the image versions, only the KiCAD project, so that's something I will do. 
For the SS input circuits, I made them duplicate hierarchical pages to indicate that they are all the same circuit and that they are fungible. I didn't want to have duplicate content in my schematic. Unfortunately, it appears KiCAD won't let you include the same hierarchical page in multiple places hence the duplicates. Is there a better way to go about this? 
Many of the other concerns have been resolved already in the current version. The relays are marked as off board and connected through the M.2 (although I do believe I forgot to add a second pin for the coil current). The references have been put back. 

I'll go ahead and make the couple of changes I actually do need to make and I'll update the PR soon.

jack-greenberg (changes requested):
This looks great! Couple things:
* I notice you using a few acronyms that are non-standard on the team, like `mobo` (I don't know what this means) and `WD` (I assume weld detect). Please spell those things out since they are non-standard and it's better to have more clarity when naming things, rather than just shortening names.
* I was thinking about the inclusion of the AIRs on your schematic this morning, and I realized that what I find confusing is that when I look at a schematic, I usually assume that every symbol will have a footprint and be on the board, as is standard in schematic capture. One way that you could still capture the connections to the AIRs is to just include a PNG of the AIRs on your schematic. There is a way in KiCad to add images, and I think that might be more clear than including symbols.
* Why are the signals for driving the AIRs going through the M.2 connector to the motherboard? I figured it'd be easier to just pass those to a Microfit connector that connects directly to the AIRs. @mkazc @elviswolcott thoughts?
* Can you add net labels to M.2 connector pins 12-13, 15-16, and 18-19 like you did for pin 10 (precharge)?
image
* You've got some errors and warnings in your ERC (electrical rule checker) in KiCad. Have Melissa show you how to run it and see if you can resolve some of the issues. Some of them are due to you adding the AIR symbol in your schematic, which I mentioned above of being something to consider changing.

Nice job otherwise!

jack-greenberg:
Also make sure to update the title of your Pull Request to something that is more appropriate! Feel free to take inspiration from other PRs that are open right now.

mkazc:
@jack-greenberg Not sure if this was communicated yet, so replying here so everything is in one place - we're having the signals go through the M.2 connector to the motherboard so that all the connections we need to come out of the same place in a wire bundle. The intention is that it will be a lot easier to deal with and route that way for both us and enclosure side. Thx!

jack-greenberg (approved):
Just update your PR name like we discussed and run the KiCad git filters, but other than that, looks good!

jack-greenberg (changes requested):
Overall, this looks great! Nice job. Just a few things:
* You have a silkscreen label called "AIR-" which is pretty vague. Consider changing it to AIR- LSD or something
* None of your shutdown nodes are labeled, which makes them not super useful. I would figure out a way to add labels
* I'd argue you don't really need a ground test point because you already have a huge exposed ground pad where we add the screw for the M.2 connection, but up to you
* You should avoid running any traces below your crystal. The signal on that wire is highly susceptible to noise, and while it *might* work, it would be better to reroute it around the crystal. Also, avoid sharing ground vias between the load capacitor and the crystal
* On the silkscreen note again-- you have "WELD DETECT" but don't show which is which (plus vs minus) which I can imagine being confusing. Can you add labels to indicate which LED is for which?
image

Let me know once you've got these changes in and I'll give it another look!

bubbleebubbles (changes requested):
Everything looks much better, thanks for implementing the feedback! The back ground plane looks much better with the shorter traces and I'm happy to see all the bypass caps with their own GND via. Couple more points of feedback for you below: 

* Is there a reason why you have two 5V test points? I'm not sure that is necessary. 
* We don't need a test point for AVCC. Which test case did you envision when you added it? 
* nitpick: The weld detect LED labels are clashing a bit with the shutdown sensing LEDs. I would consider moving "Weld Detect" to the left of the 2 LEDs where there is more space. Labelling with a colon "Weld Detect:" would make it clear enought what you're pointing to. 
* For the programming header, it would be helpful to label a least pin 1 so that we plug everything in the right orientation.
* When I ran DRC, I saw an error where the trace connected to C3 wasn't connected. Please fix before generating gerbers. You can ignore the other error since it's for the GND pad on the edge of the board which is intentional. 
* The GND copper pour in the image below isn't necessary. Remove the pour and connect each GND pad to an individual via. 
gnd_pad


Overall, everything looks great! Thanks so much for the hard work! :))

vrhinehart:
I have two 5v test points because I wanted one close to the micro for things like scoping its power input. I have the AVCC test point because I was putting test points on all power supplies and that's a separate one by way of a resistor. I don't really know what it does, but I can take it off if you want me to. My default stance is they don't eat much so I don't see a reason to get rid of them. I've implemented everything else.

vrhinehart:
I'll leave the test points. They aren't in the way. Had they been difficult to find room for, I would have put more thought into adding them in the first place.

jack-greenberg (approved):

bubbleebubbles (approved):

PR #193, opened 2022-11-28
Fix the filters install so it works on Mac and Linux both

vrhinehart:
I just needed to add a space at the end of the command.


jack-greenberg (approved):
Looks good! Didn't realize this was an issue, and thanks for bringing us one step closer to MacOS support haha.

PR #194, opened 2022-11-29
Remove duplicate files with case-sensitive names

vrhinehart:
This issue primarily affects Mac users, which I realize is unsupported, but I looked at the files and they are not needed to be duplicated. More details in commit notes.


elviswolcott (approved):
This seems like a pretty reasonable solution and seems like a safe change to make.

PR #36, opened 2021-09-17
Integrate CAN Software Update

jack-greenberg:
This pull request tracks merging of the branch `jack/btldr` into `main`. Before that, multiple PRs will be opened that will be merged into this branch, and once all those pull requests are reviewed and the system works, we will then merge this into `main`.

PR #196, opened 2022-12-02
Fix issue causing firmware images to pull in unnecessary code

jack-greenberg:
tl;dr there was a compiler flag that got lost when we merged #36, so here I am adding it back in.

## Background: Compiling

When we compile code, first, we take whatever `main` file we have and convert that into assembly, and then we _link_ that assembly in with the other libraries (like the CAN library or the GPIO library).

## The issue

When we were seeing build failures in CI, it was a failure at the linker step:

```
bazel-out/k8-opt/bin/libs/can/test/test_app.elf section `.text' will not fit in region `text'
```

This means that we are trying to compile code, but the microcontroller target we are compiling __for__ has too little flash memory to store the firmware.

This shouldn't be happening because we've never had issues with this specific image before (it's very small). I realized that if we were overflowing the flash, it means we were pulling in code that we shouldn't be. I've run into this issue before and realized that we were probably missing a compiler flag:

```
--gc-sections    Enable garbage collection of unused input sections.
```

This option looks at your code and removes any functions and data that are never used in your program. We used to have this, but it disappeared at some point in our git history. Adding it back in fixes the issues.

## How did this happen?

We never had these issues a month ago, so it is perplexing that we are seeing them now. I remember running into this issue before so I did a search through our monorepo for `gc-sections` and found that it was in [bazel/defs.bzl]:

```python
    if kwargs.get("btldr"):
        btldr = kwargs.pop("btldr")
        linkopts.append("-flto")
        linkopts.append("-Wl,--gc-sections")
        data.append(btldr + ".hex")
        defines.append("BTLDR_ID=" + ECUS[name]["btldr_id"])
```

Notice! This is all wrapped in a `if kwargs.get("btldr")`, which means that this compiler flag only gets added if the target has an associated bootloader--which *is* the case for all of our firmware images _except_ a small handful, including the CAN test image in `//libs/can/test:test_app`. This explains why it was failing but the other images weren't.

It must be that I was refactoring some commandline options when working on #36 **Integrate CAN Software Update** and the issue didn't start to rear its head until now.

[bazel/defs.bzl]: https://github.com/olin-electric-motorsports/olin-electric-motorsports/tree/619b20fb6388a114162f33084e91f4145402107f/bazel

# pytest version

Inexplicably, this then uncovered an issue with `pytest` that caused the `test` phase of our build to fail, but a quick web search [led me to a result](https://github.com/pytest-dev/pytest/discussions/9195) that fixed it! As an outcome, **this PR also bumps the version of pytest**.


ayushchakra (approved):
Looks good! Thanks for fixing this!

PR #198, opened 2022-12-03
Add roadkill harness schematic

jack-greenberg:


bubbleebubbles (changes requested):
Feedback, woooo. 
## RKH Schematic Feedback
### Buck: 
* add LEDs for power at Buck_Vin and Buck_Vout. 
* Consider fusing the output of the buck
* LED for buck output should come before the fuse so that we can test it under load 
* Optional: could add a screenshot of the table that gives you the resistor and inductor values to choose 5V for the output voltage. It's more of a readability thing. 

* everything else looks good here :) 

### General 
* The raspberry pi should be able to silk up to 16mA of current into each of its GPIO pins. We could consider just connecting the coil of the relay to the individual gpio pins and driving it that way but we're also not constrained on space so low-side driving them works well. 
* TSMS, Estops, IMD Latch relays look good! 
* Add a fuse before the GLVMS switch
* Will you have the option of connecting the actual parts we use as part of the test? For example, if we wanted to connect the inertia switch itself. Or did we already talk about this and say it's not worth it lol. 
* I think final_shutdown needs to go to the air control board. 

Everything else looks good so far!

amit-kumarh:
## Feedback
Haven't looked at the service section just yet, but have to step away so here's what I saw everywhere else.

### Main
- Should all the HITL_GPIO# labels be given a number?

### MSB
Color SHDN wire on the E-Stops purple to match the rest?

### Nostril
- BOTS is now being sensed on the Dashboard

#### Dashboard
- Connect up the 4 start button pins
- BMS and Buzzer LED pins are swapped
- On the status LEDs, there are already current limiting resistors onboard the hat; not sure if you need the ones there
- No 5V on the Dashboard M.2
- CAN P and N are switched on the (just realizing this when comparing pinouts - something we should change on the dashboard?)

#### Throttle
- Is this sheet finished? Bunch of the M.2 pins not connected and the APPS box is floating over on the right.


Overall, looks really clean and easy to understand! Thanks for all your work on this!

amit-kumarh (changes requested):
see prev comment

ayushchakra (commented):
Couple of notes/questions:
1. Would it be possible to include a label on the silkscreen for the HITL just to be consistent since all the other connectors are labelled?
2. Might be nice to indicate the orientation that the HITL and raspi need to attach just to avoid the possibility of connecting them incorrectly. . 
3. What's the reason for not being to power the HITL through the 12V bus of the RKH? Wouldn't it be easier for both to be able to operate on one power supply rather than requiring two different ones?
4. Could you add a "TERM" label to the second termination resistor to be consistent with the first one being labelled?
Other than that, the layout looks great!

jack-greenberg:
Addressed all changes from @ayushchakra.

> What's the reason for not being to power the HITL through the 12V bus of the RKH? Wouldn't it be easier for both to be able to operate on one power supply rather than requiring two different ones?

If you look at the 12V net on the HitL that goes to the connector, it is not used for powering the HitL, it is used to power any 12V connections on the board it is testing. We will need to power HitL separately.

amit-kumarh:
Sorry about the delay in the review. Looks good to me! Only things I have are this label is on the front copper layer - not sure if you wanted it there?

Jan10::220914

And there's also no name/rev number/date/logo et al. There is the QR code on it for the docs (great idea by the way!), but throwing at least the rev number/date on the back can't hurt in case we have multiple versions of this board.

amit-kumarh (approved):

jack-greenberg:
Yeah the title is on the copper layer on purpose--it looks really cool when you use ENIG as the surface finish. But now that I think about it, we might be using HASL since it's cheaper. I'll ask Elvis and if it's HASL I'll just move to silkscreen.

And good point about the rev. Also forgot to add my name lol.

PR #200, opened 2022-12-06
Add STM32G431 Dev Board

mtong1:
Based on existing firmware tester board I modified the board to be compatible with the new STM32 microcontroller.


drewnotdrew:
Picking this up.

PR #201, opened 2022-12-06
Jun/precharge discharge

redhotchili0909:
V.1 of PCB Design


bubbleebubbles (changes requested):
Great job so far! Please see below for a couple points of feedback: 
## Layout Feedback
* Add silkscreen markings to show the distance between HV and LV (see last year's board as an example). 
hv_lv_markings_needed
* Make sure traces are bent at 45 degree angles and not 90. It's not gonna break the board but a good practice to have. 
trace
* for any HV power connections , please make the traces 30 mil. Such as the traces going to and from the discharge/precharge resistors. LV traces are ok being 6 mil. 
* Can you add the version number for your board and a MKVI. (ie. Rev. 1)
* Also, on schematic, please add a pull-down resistor (10k Ohms) between the gate pin and GND for the Precharge LSD mosfet. 


Thanks for the hard work! 

*

bubbleebubbles:
Small correction on my part -- LV signals are fine as 9.84 mils (i originally thought they were 6 mil). Can you change them to a nice round number though, like 10 mil.

jack-greenberg:
One last note--your precharge LSD MOSFET is in the incorrect orientation. GND should be connected to the source--pin 2--not the drain (pin 3)
image

amit-kumarh:
One thing, this is the very tiniest of nits, but that's the old OEM logo on the front silkscreen. Maybe something we should remove/archive from the library...?

bubbleebubbles (changes requested):
Couple last points of feedback: 
* You have LV ground underneath HV signals (Bottom layer ground). Adjust the ground layer so that it's the right distance from the hv connections. 
* Change the hv/lv separation silkscreen to be in mm so that we clearly show that we have at least 12.7mm between HV and LV. 
* I'm pretty sure some parts are not far enough apart -- 476 mils is approx 12mm (we need 12.7)  FSAE rulebook shows us what the minimum distance is. See image below: 


Thanks!

elviswolcott:
I realized that we should be giving at least 2.5mm of clearance between HV traces (so we don't see shorting across relays when they are supposed to be closed). Did that and then rebased to resolve the merge conflicts. Should be ready to ship now.

jack-greenberg:
This PR adds a `jun.pretty` library. Should we merge the footprints into our main library? Or is there a way of having a footprint without it being in a library? I think there is...

jack-greenberg (dismissed):
Also, there's no mounting hole for an M.2 screw, which I think is especially important for this board as it will be quite heavy.

elviswolcott (approved):

PR #204, opened 2022-12-15
BSPD PCB Update

Eemac:
Adding PCB Layout + some minor changes


jack-greenberg:
This layout looks great! Nice job with your placement and routing. A couple things I found:
* in the BrakeLight sheet, you have a Brakelight LL label that doesn't actually go into the sheet. It's not connected in the sheet. I believe it's supposed to be connected to B_OT? But B_OT isn't a great label because I don't know what it means.
* You should add silkscreen that indicates the following:
  * Your name
  * The board name
  * The revision (rev 1 in this case)
  * The date (today's date is fine)
  * The vehicle (MKVI)
* Can you take another look at what labels are inputs vs outputs? It will help make your schematic a little clearer (for example, the BRAKE_PRESS_SENSE should be an output from the point of view of the M.2 connector since it is outputting from that component). They mostly look correct, just take one more look for me :)
image

aside from that your layout is clean AF. I'm not suuuper crazy about the amount of a gap in your ground plane (pic below), but it shouldn't be a huge issue i think. You should ask @bubbleebubbles too though. I really love the way you have your test points and LEDs together and labeled--that's a really great design choice. Great job, and lemme know when you have these things addressed and I'll re-review :)

image

jack-greenberg (changes requested):
See previous comment

jack-greenberg (approved):
This looks great! Nice job on the silkscreen especially--very clear and easy to understand.

bubbleebubbles (changes requested):
Great job on the layout so far! I'm super impressed! Please see below for some feedback. 

## BSPD Layout Feedback
* You have a stray trace here (see image) 

* The bypass capacitors for the NOR gates aren't connected to the power pins they're meant for. If they're not placed as close to the pin as possible and they're not connected directly, they don't really do much for the part. I think everywhere else is good. 

* You have a lot of traces going through the bottom layer, which is fine but there are some small/quick fixes you can do to help a little with that. I'll add a couple suggestions below but make sure to take a look around the rest of your board to see if there is anything you could do (feel free to message me if you would like help with this). I wouldn't go about optimizing for this though, your board should still work as Jack said in his review. 


* One thing that helps keep bottom traces shorter is grouping your top layer traces as much as possible. If you hover over a trace and press D, you should be able to drag it without having to redraw it. This way you can bring traces together to the minimum distance allowed. For example: 

* For less sentitive signals (not communication signals or analog signals), don't be afraid to route through multiple vias. Don't go too crazy with this but if it helps minimize the interuptions to the ground plane, go for it. 

The silkscreen looks awesome and I'm really happy overall with how clean this board looks! I'm excited to ship this board :) Let me know if anything above is confusing/if you need me to clarify anything. Good job!

amit-kumarh:
Didn't optimize the ground plane interruptions, but did switch the bypass cap pins. If y'all think its critical, can spend some more time on it, but wanted to get this to a shippable point ASAP

bubbleebubbles (approved):
That's fine, I mostly wanted ian to be aware/have the option to. Board should still work.

PR #205, opened 2022-12-15
Add rev 1 of throttle PCB layout

eys123:


bubbleebubbles (changes requested):
First pass looks great, thanks for the hard work! Leaving a some feedback below -- please reach out if you have any questions about any of it. :) 

## Throttle Layout First Pass Feedback
* Silkscreen: Your OEM silkscreen is not going to show up on the PCB when we manufacture it. To see a 3D model of the board, you can press alt-3. 
* Silkscreen: Make sure your silkscreens aren't overlapping and that you keep the silkscreen close to the part you're refering to. 


* Feel free to give each GND pin its own via down to the GND plane on the bottom of the board. That way you can reduce the length of some of your traces. We want to minimize the length of traces on the bottom of the board, so by giving R6 a gnd via near the gnd pin, you can shorten the 5V trace underneath significantly. Please check the rest of your board for things like this :) 

* Another example of how we might rearrange traces to reduce interuptions to the ground plane at the bottom: 

* Try your best to put as many traces as possible routed closely together (like the image above). That way you have more freedom to shorten the back copper traces. 
* As a general best practice, keep your traces to 45 degree angles.

* Bypass capacitors should always be placed as close as possible to the power pin of the microcontroller: 

* Silkscreen: Add your name, board revision, car version, and board name to the silkscreen before we ship.Something like this:  
Ellen
Throttle Rev 1
MKVI 2022 

This was a lot so please reach out if any of these comments are confusing! Awesome job, I'm super impressed with this first pass! :))

jack-greenberg:
image
One quick note on your schematic. Each pin--especially analog pins--should have a ground directly next to it. In this image, INPUT_THROTTLE_1 doesn't have a GND next to it (pin 9 doesn't count since it's separated by a significant gap). Can you rearrange your pins to add a GND next to INPUT_THROTTLE_1?

bubbleebubbles (approved):
Looks good to me!

jack-greenberg (changes requested):
Layout looks good, but there are a ton of merge conflicts in the schematic file :(

jack-greenberg (approved):

PR #206, opened 2022-12-22
Mkvi bms layout

mkazc:
Status:

- **DA Module: Rev 1 Done** - Finished recommended edits below, may change silkscreens
- **BMS CSC: Rev 1 Done** - Formerly named the Peripheral, name changed to match purpose and standards. Finished edits. Keeping transistor as is, and will include small heatsink. Going to write more on that on Coda page.
- **BMS Core: Rev 1 Done** - Made all recommended edits.


jack-greenberg:
Layout looks great for the DA module. Only thing is that your GND mounting hole is rotated. The white line should line up with your board edge.

image

bubbleebubbles (dismissed):
Hihi, great job on the thermistor DA board! Gonna leave a few comments below :) 

## Thermistor DA Layout Feedback: 
* The copper pours in these two regions (see image) aren't necessary since kicad is just drawing the traces instead of adding the pour. Remove the pours and add thick traces between the Vref pads. You should be able to get away with 20-30 mils and it'll look cleaner and be better for manufacturing. 

* nit (but generally a good practice/habit), make sure your traces always come in/out from the center of the pad. (see image below) 

* nit. I would prefer to have the thermistor 1-8, 9-16, 17-24 a bit closer to the chips to make it clearer what you're referring to. There's lots of space on top of the traces for silkscreen. You could also always add a silkscreen box around each section and add the label right underneath to be as clear a possible. 
* It would also be useful to add labels for thermistor 1 and thermistor 24 near the connector so that we know which thermistor is on which side of the connector. 
* Should this point towards the trace you're refering to just like the rest of the board? 

* I missed this earlier. You're missing a bypass cap on the MCP6001 chip. [datasheet](https://ww1.microchip.com/downloads/en/DeviceDoc/MCP6001-1R-1U-2-4-1-MHz-Low-Power-Op-Amp-DS20001733L.pdf)


* Add a silkcreen dot to the LTC chips for where pin 1 is, in case we ever need to depopulate and put them back on. 
* Your I2C lines are missing pull-up resistors. Take a note on the peripheral board to add pullup resistors on the SDA and SCL lines (4.7k or 10k). If you have time or space, it would be useful to add a placeholder on the DA board/DNP pullup in case future peripheral boards miss those. Alternatively, you can add some copper pads to add resistors to later, to be safe. Since the DA boards are not likely to be redesigned, it would probably make sense to add placeholders AND add the resistors on the peripheral as well. 
* Your I2C traces look ok -- if you can length match them as much as possible that would be great but I don't expect to see any problems with the way they are routed right now. More of a note for the future in case you have to change things. 

Overall, this looks great! Thanks for the hard work! <33

jack-greenberg:
# Peripheral feedback

Overall, this looks really great. Great use of 2D space and impressive given the constraints. Most of my feedback will be nit-picky and a lot of advice I will give is stuff I learned recently working with Brad for my capstone, so I'm excited to share some info :)

# Schematic

* Remember that for the I2C addresses you either need to mark one/both as DNP _or_ (I would recommend) use a 3-way solder-bridge and you can just jumper it with solder

# Layout

* IC1.10 (IC1, pin 10) is not connected to GND
* U1.17 is not connected to its net
* R49 and R50 have overlapping courtyards
* Unless it is justifiable, I would avoid thermally dissipating Q1.4 on the bottom layer and instead keep it as a GND plane. I'm not sure you need the second plane and it could be beneficial for power integrity to keep a solid reference below the copper pour on layer one. Not totally sure how it will perform, but maybe we can discuss it more or have Brad weigh in
* You are only using one via size throughout, and there are some areas where you can probably get away with smaller vias. As a rule of thumb, I usually use 3 sizes (total size/drill): .7/.3mm, .6/.25mm, and .5/.2mm. You have vias (like below) that could be smaller and would be totally fine and take out less of your GND plane. Not a huge deal, but something to consider
image
* I would place vias for your EPAD on U1 like so (below). This is what I usually see recommended (you might want to arrange them in a slightly nicer grid than I did--that was just for show :smiley: 
image
* In general, your vias can be placed closer to your pads, like the images below. When using GND vias, it's recommended to use a wide trace (like almost as wide as the pad, where appropriate, 10-15mils) to a via right on the pad to minimize current loops.
image
image
* I would place a GND via right next to C1.1 even though you have it going to a GND plane. Think about the path of least resistance for an electron to flow back to the producer of the signal.
* I think 3D models are missing for the isoSPI connectors
* Don't forget to add test points for ADBMS/multiplexer! I recommend small ones like little circular pads with .75 or 1mm diameter
* Maybe the new OEM logo? :)

---

Overall, again--incredible job. The fact that you only have a single signal going through a via to the bottom plane is really astounding. Really impressed by your progress and let me know once these changes are in and I can take another look!

jack-greenberg:
For peripheral: I strongly recommend doing a GND fill on B.Cu so that you have a reference for your 5V power plane, otherwise you get a lot of electric field spread which can cause EMI issues. Instead of routing traces on B.Cu, route them on PWR.Cu. It's better to have cuts in your power plane than it is a GND plane. Let me know if that doesn't make sense.

jack-greenberg:
One more small thing for the BMS Core (sorry :sweat_smile:): The isoSPI connector overlaps with the M.2 mounting hole. I imagine there will be a clearance issue because we won't be able to fit the screw there. Can the connector be moved north at all?

jack-greenberg:
Sorry for accidentally closing :sweat: 'twas a mistake.

elviswolcott (approved):
Based on prior reviews and looking through everything, I think we good to go on the BMS!

PR #207, opened 2022-12-23
Remove unused 3D models

jack-greenberg:
I noticed that our monorepo is getting pretty chonky, so I wrote a script that went through our 3D Models and deleted all the ones that we aren't using anywhere. The script is here:

```bash
#!/bin/bash

for file in parts/3DModels/**/*;
do
    FILENAME=$(basename $file)
    INPATHS=$(git grep ${FILENAME})
    if [ -z "${INPATHS}" ]
    then
        echo "${FILENAME} is unused"
        rm $file
    fi
done
```

Effectively, we are iterating over each file in 3D models (and in each sub directory in it) and if the file isn't mentioned anywhere in the monorepo, it gets deleted. If a 3D model is used, it will be contained in the footprint file or in the .kicad_pcb file, so if it isn't found anywhere (using `git grep`), then we delete it.

This takes the parts/ directory from 246MB to 108MB.

Please please double check me on this--I could've made a mistake that might be annoying to fix :sweat_smile:


mkazc:
I'd say to keep the Microfits since a couple boards are using them this year and they may be nice to have in the future, but other than that it looks like it worked correctly :)

jack-greenberg:
Ah! One thing to clarify: some of these were duplicates. So we had two different 3D models for different components with different formats. So like a WRL file and a STEP file. Mostly we just use STEP. So we still should have the microfit 3D models.

mkazc:
Ah yeah that checks out, by chance I think I caught it removing the main file step file for ie. `MicroFit-24-RA` by accident? Though in that case it might've just been unused

jack-greenberg:
Yeah it looks like we don't have a footprint in our library for Microfit 24 RA.
image

elviswolcott (approved):

PR #209, opened 2022-12-29
Add MKVI Nostril Motherboard

amit-kumarh:
Adding the schematic for the nostril motherboard.

Biggest question I have here is which components should get power/ground from the motherboard and which will be powered elsewhere?

For this version, I'm assuming the Dashboard LEDs and RTD Buzzer will get power from this board, while the brakelight, current sensor, and brake pressure sensor will get power elsewhere, but there are pins to spare to change that.

also this is still missing the sensing controller

**Rev 2**
Layout ready for review!

Pending todos:
- Double check sensing board clearance (board footprint changed last night)

**Rev 3**
- Changed CAN Logger Microfit type and added silkscreen
- Fillet them corners
- Add mounting holes on corners and M.2 boards
- Put Docs QR on the front
- Add board outlines to silkscreen
- Add RA 23 pin ampseal 3D Models


jack-greenberg:
Looks good! Sorry for the delay in review. Couple notes:

* For dashboard pinout (this goes for the dashboard PR too #181) pins 34-37 don't follow the GND-SIG-SIG-GND pattern. Probably not a huge deal but we should be consistent
* Not a huge deal, but you established a pattern of using M.2 pins P1-3 as 5V and P38-40 as GND, but throttle uses a different pattern. Again, not a huge deal, but maybe give it a little thought
* Your buck converter is missing fusing. You should fuse the Batt+ _and_ +5V nets
* Add some LEDs for your buck converter for power indication.
* Add test points! Lots and lots of them so that we can use this MoBo to debug things if need be.

To answer your questions:
* Your board should provide power and ground for the brake pressure sensor. The current sensor is in the accumulator, and the brakelight can get +12V directly from the battery or the accumulator ampseal

Don't forget to annotate your schematic too! Nice work! Looking forward to seeing the layout.

jack-greenberg (changes requested):
This looks so good! Great job. A few notes:
* Can you make the microfit MicroFit_VT_4?
* Can you add silkscreen indicating the the CAN logger should be plugged into the MicroFit 4-pin connector?
* I recommend adding a radius to the edges of your board (round the corners)
* Maybe for your mounting holes, use a MountingHole footprint, which has a ring of copper. Looks nicer and helps prevent wear on the PCB. Not critical though, so feel free to ignore.
* Maybe put the docs QR code on the front so that people can scan it when opening the nostril?
* Make sure to add M.2 standoff mounting holes.
* You don't need to, but it'd be cute if you added the board outlines to the silkscreen of the board...
* Ampseals aren't flush with the edge of the board. Is that okay?
* Add 3D models for ampseals so that we can put it in CAD
* Make sure that with your M.2 connector placements, there is enough room to plug the boards in. It looks like it miiight be okay, but I'm not 100% sure
* Can you move the CAN test points elsewhere so that you don't create an extra-long branch on the bus? Not critical, but good practice
* Aside from that, looks great! Nice work Amit.

jack-greenberg (approved):

PR #210, opened 2022-12-29
Add BSPD SS to Brakes Microcontroller

amit-kumarh:
Very tiny PR to add one element to the brakes board that was missing, which is routing the BSPD shutdown sense signal to the microcontroller.

Although it already has the direct output of the BSPD circuit, for the sake of consistency with the way we monitor the shutdown circuit as well as the certainty that we are actually measuring the behavior of the shutdown circuit itself when we are debugging it, it is something that would be prudent to add to the board.


jack-greenberg (approved):
Good catch! Thanks for fixing that.

PR #211, opened 2023-01-03
Symbol Library Converter Tool

elviswolcott:
This PR adds a tool (`//tools/symbols:convert`) that contains two commands.
1. `symbols2library`: converts a directory full of files containing individual symbols and combines them into one `.kicad_sym` library file
1. `library2symbols`: converts a `.kicad_sym` library file into individual symbol files

This serves two purposes. First of all, it makes reviewing changes to the parts library easier: it should be more apparent what symbols have been modified and which ones have been edited without looking through a huge single file. Most importantly, it should make merge conflicts within the parts library less frequent and make it easier to resolve them.

I have also added a script that installs [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) to run the script. When you checkout a branch the hook will automatically generate the library file based on the symbol files in git and when you make a commit it will automatically convert the library files back into symbols to store in git.


elviswolcott:
incorporated to #217

PR #215, opened 2023-01-23
Complete rev 2 of IMU schematic

drewnotdrew:
Revision 2 schematic of the IMU. 

Changes from revision 1:
- 12V->5V LDO instead of buck
- SOT23 LDO instead of 4-XDFN (for powering IMU)
- THT holes for direct wire-to-board connection (planning for external wire-to-wire connector)
- Tag-Connect instead of headers
- Logic level shifter between MCU (5V logic) and IMU (1V8 logic)
- Programming LEDs decreased from 3 to 1 for space
- Solder pads w/pull-up resistors for setting IMU ID externally
- Solder pad in series with the CAN termination resistor

Documentation on this project's [Coda page](https://coda.io/d/Documentation_dbuFnC2EA_e/IMU-Rev-2_su_g3#_lu0jv) provides context for some of these changes. It's still a WIP though.


jack-greenberg:
Nice work! A few review notes:
* You're missing a bypass capacitor on pin 13 of the IMU. Double check the datasheet to see what value you need
* You might consider connecting the *ERROR pin of the 1v8 LDO to the microcontroller. That way if something goes wrong your microcontroller can detect it.
* Is there a reason you have your programming LED's GND connected to the micro? It might make it harder to route because then you need to route two wires (from 5V to the LED, and from the LED to GND). I'd recommend instead connecting the resistor to the microcontroller's pin and then the cathode of the LED to GND
* You should add bypass capacitors to the VCCA and VCCB pins of the level shifter. 0.1uF should be fine
* Love the inclusion of tag-connect, but it isn't something we really support on our team yet. Maybe we should switch back to a 6-pin header for now until we have better support. Otherwise we run the risk of not being able to program your board! But also, on the tag connect, there's no need to include CAN_TX and CAN_RX since we don't use those signals for debugging--we only use CAN_+ and CAN_-
* Consider adding test points for the SPI between the microcontroller and the IMU, probably on the 5V side (or maybe both? if the test points are small enough)

elviswolcott (approved):
rebased and force pushed to resolve merge conflicts - looks good to merge!

PR #216, opened 2023-01-26
Add in Tunables Parameter Library 

tlee10333:
Python Code Only for Now.


tlee10333:
I wrote tunables_shell.py, which creates a tunables shell environment which does exactly what tunables_interface.py did. Also there's the addition of a function called list which reads the yaml file and searches and prints the specified message.

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,113 @@
...
+        try:
+            data = yaml.safe_load(file)
+            for i in range(len(data)):
+                for j in range(len(data[i]["params"])):
+
+                    message = data[i]["params"][j]
+
+                    if message["name"] == arg[0]:
+                        if message["edit"] == False:
+                            print(f" {arg[0]} cannot be edited")
```
ayushchakra, line 94: I feel like you should be raising an error here, makes it more clear to the user that what they are doing can't be completed

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,113 @@
...
+                            break
+
+                        data[i]["params"][j]["current_value"] = literal_eval(arg[1])
+                        print(date.today())
```
ayushchakra, line 98: Maybe add a string before this saying what is being printed, could be confusing to the user to just see the date being outputted to them

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,113 @@
...
+                        data[i]["params"][j]["date_modified"] = date.today()
+
+                        break
+            file.close()
+
+            with open("tunables.yml", "w") as file:
+                yaml.dump(data, file, sort_keys=False)
+                file.close()
+
```
ayushchakra, line 107: Will this eventually also include the workflow for sending values via CAN to the target mcu?

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,113 @@
...
+# import database
+
+
+class tunables(cmd.Cmd):
+    print(
+        " \n This is the Tunables Parameters program. \n Type in ? to see the list of commands"
+    )
+    prompt = "(tunables) "
+    file = None
```
ayushchakra, line 16: Based on how you're parsing the user's input, it seems that you are expecting a certain format for the data to come in, might be helpful to prompt the user about what format you are expecting just to avoid misuse or incorrectly passed in parameters

ayushchakra (commented):
overall this looks great! can't wait to see you start writing the C portion of this :)

also side not but don't forget to lint your code

libs/tunables/test_recieving.yml
```diff
@@ -0,0 +1,25 @@
...
+             f: OFF
+
+subscribe:
+  - name: dashboard
+    mob: 1
+
+
+
+
+
```
jack-greenberg, line 25: Extra newlines

libs/tunables/tester.c
```diff
@@ -0,0 +1,23 @@
+#include "libs/tunables/can_api.h"
+
+/*
+
```
jack-greenberg, line 4: Extra newlines

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+    GETTER = 0,
+    SETTER = 1,
+    MEASURE = 2
+}
+
+// A global variable that gives each board maximum of 16 tunable parameters
+uint32_t TUNABLES_MEM[16] __attribute__((section(".eeprom")))
+= { 0 };
+
+// Can't I just declare send_can in tunables_init??????????????????????????
```
jack-greenberg, line 15: `send_can` is a global variable that must be declared in the global scope as:

```c
static volatile bool send_can = false;
```

I can explain what `static` and `volatile` mean another time, or you can look them up!

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+void timer0_callback(void) {
+    send_can = true;
+}
+
+uint8_t can_get_data[2] = { 0, 0 }; // Data that we received???????????????????
```
jack-greenberg, line 20: Yes, this is the spot for the memory that we will receive.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+        .dlc = 2,
+        .data = can_get_data,
+        .mob = 4,
+    };
+
+    // can filter can_get_filter
+    can_filter_t can_get_filter = {
+        .id = ecu_id,
+        .mask = 0x7FF,
+    };
```
jack-greenberg, line 38: These variables will need to be declared globally. In C, you cannot create a variable in a function that you intend to be used elsewhere.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+void tunables_loop() {
+    // Is rc going to be a local variable??????????????????????????????
```
jack-greenberg, line 42: Yes.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+    int rc = can_poll_receive(&can_get);
+
+    // If there was a can message
+    if (rc == 0) {
+        // Just to clarify, tunable_id is declared in the tunables.yaml
+        // file??????????
```
jack-greenberg, line 48: Yes

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+
+        // can_get
+        if (data[0] == 0) {
+            eeprom_read_dword(&TUNABLES_MEM + (4 * tunable_id))
```
jack-greenberg, line 52: I believe this function returns the memory, so you will need to assign it to a variable.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+    // Is rc going to be a local variable??????????????????????????????
+    int rc = can_poll_receive(&can_get);
+
+    // If there was a can message
+    if (rc == 0) {
+        // Just to clarify, tunable_id is declared in the tunables.yaml
+        // file??????????
+
+        // can_get
+        if (data[0] == 0) {
```
jack-greenberg, line 51: Instead of 0 and 1, you should use the enums you declared.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+        if (data[0] == 1) {
+            eeprom_write_dword(&TUNABLES_MEM + (4 * tunable_id), data[2])
+        };
+
+        can_receive(&can_get, can_get_filter);
+    }
+
+    // Start listening again!
+    if (send_can) {
+        can_send_test_msg(); // what does this do???????????????????
```
jack-greenberg, line 65: This sends a test CAN message that you can use for debugging purposes. Can be removed.

libs/tunables/tunables.c
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// A global variable that gives each board maximum of 16 tunable parameters
+uint32_t TUNABLES_MEM[16] __attribute__((section(".eeprom")))
+= { 0 };
+
+// Can't I just declare send_can in tunables_init??????????????????????????
+void timer0_callback(void) {
+    send_can = true;
+}
```
jack-greenberg, line 18: You don't need this. It will be declared in the application file. this is a library file.

libs/tunables/tunables.c
jack-greenberg: Also though you don't need it. see next comment.

jack-greenberg (changes requested):
Please also run the linting tools:

```
./scripts/clang-format.sh
bazel run //:buildifier
```

And there will be changes so commit those changes.

libs/tunables/templates/h_file.j2
jack-greenberg, line 1: So you are also going to need a c_file.j2. Currently, you're declaring a variable in a header file, which can work but only in specific, basic examples. In the current file, you will need to just declare the type of the variable:

```struct {{ name }}_tunables { // name will be populated by like "throttle" or "bms"

};

extern struct {{ name }}_tunables tunables; // Declares the variable to be shared
```

In the C file, you'll need to declare the variable itself:

```c
struct {{ name }}_tunables tunables = { 0 };
```

jack-greenberg (changes requested):
Haven't looked at python yet, just left some initial comments about the need for another template file.

libs/tunables/BUILD
```diff
@@ -0,0 +1,46 @@
...
+#         "//bazel/constraints:atmega16m1",
+#     ],
+#     visibility =
+#         [
+#             "//visibility:public",
+#         ],
+#     deps = [
+#         ":can_api",
+#     ],
+# )
```
jack-greenberg, line 46: Remove

libs/tunables/can_layer.py
```diff
@@ -0,0 +1,79 @@
...
+    which tunable uses to communicate between the host computer and the car.
+
+    Attributes:
+        bus: A can.Bus instance which is sets the settings for
+        which channel and bitrate we will be sending CAN messages
+        from.
+
+    """
+
+    def __init__(self, channel_para="can0", bitrate_para=500000):
```
jack-greenberg, line 19: What is `_para`? I would remove it if you don't need it. Seems redundant.

libs/tunables/can_layer.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Basic Format
+        message = [func_type, parameter_id]
+
+        # Setter
+        if func_type == 1:
+            message += new_value
+
+        msg = can.Message(arbitration_id=arbit_id, data=message, is_extended_id=False)
+        self.bus.send(msg)
```
jack-greenberg, line 60: If `func_type` is 1, it's just adding a single member to the array. If the value is greater than 8 bits, will the array get split up into more bytes? It seems like the datalength of this will always be three, so it won't include the full updated value.

libs/tunables/h_generator.py
```diff
@@ -0,0 +1,84 @@
...
+
+# I'm currently running h_generator in the tunables folder.
+# #If you're in a different folder, don't forget to input the entire path
+paramslist = get_params("bms", "tunables.yml")
+
+# Generates a C struct with all the parameters in the bms board
+content = struct_maker(paramslist)
+
+# Shows what we made via the jinja template.
+print(content)
```
jack-greenberg, line 84: At the end of this file, you should have something like:
```python
if __name__ == "__main__":
    main()
```

I understand this is just for testing. And we chatting about argparse and moving to `sys.argv` instead.

libs/tunables/tester.c
jack-greenberg, line 1: File unused, delete?

jack-greenberg (changes requested):
More thorough review.

libs/tunables/CAN_layer.py
```diff
@@ -0,0 +1,30 @@
+# Cantools used to read DBC files
+
+
+# Uses can to communicate via CAN from computer to car
+import cantools
+import can
+from pprint import pprint
```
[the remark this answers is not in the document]
tlee10333: I completely revamped the CAN_layer.py file by making a new file called can_layer.py. I'll delete the CAN_layer.py in the next commit.

libs/tunables/CAN_layer.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+# CAN section
+
+
+class TunablesCAN:
+    def __init__(self):
+        self.bus = can.Bus(bustype="slcan", channel="/dev/ttyACM0", bitrate=500000)
```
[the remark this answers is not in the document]
tlee10333: Done!

libs/tunables/CAN_layer.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def send(self, funcType, parameter_id, new_value=0):
+        message = [funcType, parameter_id, 0, 0, new_value]
```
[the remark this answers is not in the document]
tlee10333: In the Coda document for Tunables I outline how the data for the CAN msgs will be sent.

libs/tunables/CAN_layer.py
```diff
@@ -0,0 +1,30 @@
...
+
+        msg = can.Message(arbitration_id=0x6E0, data=message, is_extended_id=False)
```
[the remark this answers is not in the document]
tlee10333: It's outlined in the Coda doc

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,113 @@
...
+
+def search_yaml(arg):
+
+    """Finds information of 1 parameter"""
+    with open("tunables.yml", "r") as file:
+        try:
+            data = yaml.safe_load(file)
+
+            for i in range(len(data)):
+                for j in range(len(data[i]["params"])):
```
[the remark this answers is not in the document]
tlee10333: Currently the way the yaml file is formatted we don't use a dictionary so it's hard to search by keyword. Also since each board will have max 16~ tunable parameters, I think the for loop is a heuristic solution for now.

libs/tunables/tunables_shell.py
```diff
@@ -0,0 +1,201 @@
+"""The python shell interface that users interact with"""
+
+import cmd
+from ast import literal_eval
+from datetime import date
+import can_layer as CAN_layer
```
jack-greenberg, line 6: Why do you have the import like this? Why not just do `from can_layer import TunablesCAN`?

jack-greenberg (changes requested):
Looks good! Just a few small comments. Almost ready to stitch all this together and test!

libs/tunables/c_generator.py
```diff
@@ -0,0 +1,123 @@
...
+                file.close()
+
+            print(f"\n ########## {ecu} tunables.h file ########## \n")
+            with open(h_out, "r") as file:
+                print(file.read())
+                file.close()
+
+        else:
+            # Error msg
+            print(f"ERROR: tunables.yml at {ymlpath} does not exist")
```
tlee10333, line 59: Didn't raise an exception because I don't want to break the for loop. 

For example, the line:

python c_generator.py throttle fakeECU bms

This should generate the .c and .h files for throttle and bms but since fakeECU isn't a real ECU, it'll just notify the user and then move on.

jack-greenberg: I would argue that this is not good behavior for a program. In my strong opinion, programs that run automatically (i.e. not directly by the user, as is the case with your code) should not be fault tolerant, and instead should fail loudly and explicitly when there is an error. This is so that your code can point to issues elsewhere. For example, if your code silently fails and moves on, we lose the ability to detect if the thing that called your code (i.e. the Bazel build system) has some error or misconfiguration.

libs/tunables/api.h
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright 2023 Olin Electric Motorsports
+ */
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+#include <string.h>
+
+#define MAX_NUM_PARAMS (16)
+
+typedef enum { GET, SET, MEASURE } message_type_t;
```
jack-greenberg, line 11: This isn't needed here since it's only used in your tunables.c file.

jack-greenberg (changes requested):
Looks great! Just fix the one thing, and then I'll re-review the Python code one last time, then we can work on integration with Bazel.

Make sure to run the formatter on your code.

libs/tunables/api.h
tlee10333: Fixed

jack-greenberg (approved):

ayushchakra:
from debugging on 2/5: need to import <stddef> for size_t, would be nice to automate adding automation for defining the ECU_ID, also might be nice to automate the output directory from `c_generator.py`

PR #218, opened 2023-02-04
Add MKVI sensing controller board

ssirroon:


amit-kumarh:
Good job! Here's a few notes.

## Schematic
- In the filtering/amplifying sheets, the CH1_OUT...CH4_OUT pins are marked as inputs, when they should be outputs
- Also on that sheet, can you fill in the gain/cutoff values?
- if you want, it looks like you have a little room and potentially add another programming LED if you want to. Absolutely not necessary though.
- KiCAD is also telling me your schematic is not fully annotated.
- Question that I had, what is the purpose of C7? If its a bypass capacitor, I believe it should be connected in parallel with the chip power to GND, but the AD8630 datasheet also says no external capacitor required, so just wondering that

## Layout
- You have some unconnected nets on the board - check DRC for full details, mostly in M.2 connector and the 16M1
- M.2 mounting hole is missing
- Lots of tiny, some unconnected traces hanging around the `ws2_Vout` via here
tiny_traces png_Feb16::112148
- Adding some PWR/GND/CAN(?) test points may not be a bad idea if space allows
- Some reference designators are still present on the silkscreen (U2, U4, J2)
- Make sure to add author name(s), date, rev number, and OEM logo to the silkscreen.

Thanks again for your work on this! Hyped to see sensing going on the car :)

amit-kumarh (changes requested):
see prev comment

jack-greenberg (changes requested):
Amit caught most of what I was going to cover. A few more things:

* Make sure to delete the `#auto_saved_files#` file.
* Parts should be added to the `oem.kicad_sym` library instead of the `formula.kicad_sym` library. @elviswolcott can tell you more and help out.
* You have sensing_controller_front and _rear, but there should only be one, right?
* Not sure why you are passing the wheel-speed sensor through a gain-stage. It is a digital signal (0 or VCC) so it doesn't need gain. A low-pass filter is fine. I would also add a pull-up resistor but mark it as DNP (do-not-populate) because it might not be needed. I can explain more in person
* What are your gain values for the op-amp? What values are you expecting from the suspension travel sensor? I'm imagining that at rest, the sensor will read some voltage X and then when the car is in motion there will be a range of +/- Y from X (so your min voltage will be X-Y, max will be X+Y), so you will need to know what values you expect so you can set the gain appropriately.
* CI is failing because you don't have a BUILD file. I can help you add that
* You're using non-standard trace widths in metric units. Most of your traces should be 6 mils, except +5V and GND should be 10 mils

jack-greenberg:
Looking good! A couple quick things:
* You have R9 and R10 marked as DNP, which means that it won't get populated. I might recommend replacing it with just a 0-ohm resistor and leaving a note that it can be replaced with a different resistor for filtering. Though I don't think we'll need it...
* Your pull-ups R12 and R13 are still R_1000. Change this to `R_1K`
* Run DRC--you have some issues
* Q1 silkscreen is off-board. Either delete it or move it
* Your mounting hole is off to the side, but it should be centered with the M.2 edge connector

amit-kumarh (approved):
Approving PR - changed PCB color to black per our earlier conversation today @richardli03.

Only minor nit I have is that a GND TP may be useful as it becomes much more difficult to probe the rest of the TPs without a GND.

drewnotdrew (approved):
Also approving this PR - I added fillets to the corners which should wrap up the last of Amit's suggestions.

I do have a couple of smalls suggestions for future layout though.
### Trace Intersections
When routing trace intersections, it's usually best to make a T junction instead of a Y (with a 45 degree angle). Supposedly this was a bigger issue ~30 year ago when acid could get trapped in the acute angle, but it's not as much of an issue now. There's still EMI and thermal concerns that make it better to make T junctions. You can read more [here](https://www.royalcircuits.com/2020/07/22/trace-intersections/).
So instead of this:
image
Route like this:
image
I also think it looks a bit better aesthetically, but that's much less important.

### Vias near pads
I was under the assumption that vias close to connecting pads was detrimental, but a few resources pointed to that it's an accepted practice. If this board is manufactured by a cheaper board house, it might be worth moving the vias further away from the pads, especially since you have the space. Happy to be wrong here though - maybe it not as big of an issue as I thought!
image

PR #219, opened 2023-02-12
Add Mock Inverter

drewnotdrew:
Revision 1 schematic of the Mock Inverter.

Features:
- 440uF DC bus capacitors to mock the PX100DX
- Voltage divider and ADC to measure TS voltage as the motor controller would measure it
- SPI and 5V isolation for the ADC between LV and HV
- 6 pin TagConnect
- 4 digit 7-segment display to visualize torque requests (3 digits + 1 digit for direction)
- Corresponding 7-segment display driver communicating over SPI
- 1A 5V SOT23 LDO for the 7-segment displays (they draw about 0.5A)
- Fused +BATT, +5V, and +TS rails (PTC on +BATT and 5V, replaceable ceramic on +TS)

Documentation for component decisions is available on this project's [Coda page](https://coda.io/d/Documentation_dbuFnC2EA_e/Mock-Inverter_suHLZ#_lu5CB). In addition, all part numbers should be up to date on the schematic symbols.

This schematic is significantly more involved than the IMU, so I'd appreciate any and all feedback on the how to improve the schematic.


drewnotdrew:
Changes since the PR was submitted:
- Bleed resistors rated for 200W, in series with a normally closed relay. When LV power is received, only then does the relay open.
- Bleed resistors are two 100W rated 500 Ohm resistors, which bleed the capacitors in around 2.2 seconds.
- TS power LED, for displaying when TS power is connected and the charge in the capacitors.
- DC bus capacitors are in parallel with the bleed resistors and capacitor charge status led. This means that if the relay fails open, the TS power LED will still display the charge in the capacitors.

As always, these decisions are documented in this project's [Coda page](https://coda.io/d/Documentation_dbuFnC2EA_e/Mock-Inverter_suHLZ#_lu5CB).

skachurolin (commented):
All the comments are left in the document linked below: 
https://docs.google.com/document/d/1m7erDQbNbSjB9la3mPvMDdWJRLvH0ISXUzeOthO0fDA/edit

drewnotdrew:
Re: Review from @skachurolin: https://docs.google.com/document/d/1m7erDQbNbSjB9la3mPvMDdWJRLvH0ISXUzeOthO0fDA/edit
1. HV/LV separation: The shortest distance between HV and LV is here, with a distance of 2.6mm.
image
I found a pretty good resource [here](https://www.smps.us/pcbtracespacing.html). In short, the prescribed values by FSAE rules are much larger than what is needed for "functional isolation." Based on a few standards, clearance for functional circuits at 400VDC is roughly 2.5mm. Given that the blue ground plane is on the back and the -TS trace is on the top, the distance with a board of thickness 1.6mm is ~3mm. I'd argue that this is good enough, especially since the Mock Inverter has exposed TS.

2. Fixed.
image

3. This is a good point - I'll keep the routing through the test points but will keep this in mind for non-THT pads/test points.

4. Another good point. Fixed.
image

5. Moved all HV components after the fuse. Updated the schematic and layout accordingly. 

6. The max current through any HV traces is 0.404A, which only requires a trace width of <6 mils for a 10C temp rise with a trace 25mm long. All TS lines are 25 mils.

7. No. Originally unfused TS was to handle the inrush current from charging the large capacitors, but with precharge/discharge the current it limited enough where all HV components could be put behind the fuse. A warning message on the silkscreen was added to advise to not connect TS without proper current limiting (e.g. precharge/discharge). If TS is connected without current limiting, the HV fuse should blow.

skachurolin (approved):

PR #220, opened 2023-02-12
Add freemeter schematic

skachurolin:


bubbleebubbles (changes requested):
* I really like the top-level of the schematic! Great job :) 
* Is there a reason you don’t have a programming LED connected to the microcontroller? I find it really useful to have one but arguments could be made against it. 
* I see you’re using GND1 and GND2 for HV GND and LV GND, which is fine. We’ve had issues in the past with using these labels and it would be clearer to switch it to HV_GND and LV_GND. It’s far easier to catch a gnd mistake that way and may save us a headache in the long run. 
* It would be good to make a note of how you’re gonna connected the shunt to the rest of the system, ie. where on the car (location of the emeter). 
* The EN pin on the voltage reference U701 isn’t connected to anything so the device is neither in normal operation mode nor in shutdown/low power mode (the pin is floating). 
* Flyback looks great! 
* Could you write somewhere on the schematic, probably close to the shunt or in the current shunt page, what the expected current range that you’re measuring is. It’s’ a good reference to have to double check the current shunt schematic specifics.  
* Would also be good to note down on that page what your measurement precision is (how accurately can  you measure the current -- nearest Amp? Nearest milliamp?) 

Overall, it looks great! :)) Awesome work!

skachurolin:
Thank you for all the suggestions! I made a couple of changes and responded to most things on this document: 
https://docs.google.com/document/d/1Ap4CRJT9Jp0e7x5hu6vJoKR969G4E5jAOjC3I_JuyL4/edit?usp=sharing 

Please let me know if there is anything else that requires change!

amit-kumarh:
# Schematic
- yes, debug LED = programming LED. should be all good there
- may want a fuse somehwere in your LDO setup
- not a huge thing, but we prefer to use +Batt or similar vs 12V, since the batteries aren't exactly/always at 12V
- looks like on the top on your GND label and two connectors, you have a GND1 net label there. can we change that to HV_GND as well?
- if it's not absolutely necessary, I would consider removing the image from the schematic, if for no other reason than it makes dealing with the file a lil painful, especially if we ever get a merge conflict or something like that.

# Layout
- running DRC tells me that the opamp for the voltage divider has a GND via that is too small for our design constraints (min 12 mils)
- footprint for the optoisolator seems to have a courtyard issue
- I know you're pressed for space, so feel free to ignore, but some test points at critical points might make debugging a little easier.
 - check [this](https://olinformula.slack.com/archives/C1TLR0J92/p1669582018001919?thread_ts=1669580426.133399&cid=C1TLR0J92) out for an idea... 
- typically we don't make our boards with reference designators on the silkscreen anymore; [this](https://github.com/openscopeproject/InteractiveHtmlBom) plugin tends to make assembly much easier and kicad is usually open when debugging anyway (plus I think on this board they'd be so small that JLC's shitty silkscreen printer would make it illegible; I believe it's under 1mm rn)
- aesthetics only thing: if you like, you can also round off your corners a little bit so the final board doesn't have those sharp edges

jack-greenberg:
Haha it's been a while since I've looked at this, and it looks awesome! A few notes:

# Schematic
* On the **Microcontroller** page, the two `MISO` nets are linked but shouldn't be (you can check this by clicking on one and pressing the \` key. This means that your two SPI buses will be connected. Now, this is most likely not an issue the MISO for programming the microcontroller is only active when the chip is in reset, but still it is better practice to have them separate. I recommend naming them something like `MISO_PROGRAMMER` and `MISO_ADC` or something like that.
image
* Here, you have `+5_HV` as the name for both of the nets. This makes it so that when you do routing, KiCad will want you to connect them. But they are on different boards. In hindsight, it probably would've made more sense to just have two separate KiCad projects (i.e. freemeter_lv and freemeter_hv) but I don't think it is worth the time to make that change now. So for now, make sure that everything on the left side of the schematic has a different name than everything on the right side. There can't be anything with the same name.
  * Now that I think about it, JLC (or whoever the manufacturer is) might be upset that we have two boards in one file. There's a chance you actually will have to separate them... @mkazc @bubbleebubbles @amit-kumarh @elviswolcott thoughts?
image
* Nit
image
* Currently you have R801 and R803 as 0603 resistors. The voltage drop across these will be quite high, so can you verify that we have resistors that are rated for +600V? And if not, you need to find some resistors that are.

# Layout
* This area scares me a bit. That's a very high voltage near a much lower voltage. Maybe that's fine though? Try to get some others' opinions.
image

# General
* Remove the following files
  * `report.txt`
  * `freemeter.xml`
  * `e_meter_diagram .drawio`

Nice work :)

amit-kumarh:
>Now that I think about it, JLC (or whoever the manufacturer is) might be upset that we have two boards in one file. There's a chance you actually will have to separate them... @mkazc @bubbleebubbles @amit-kumarh @elviswolcott thoughts?

[this](https://jlcpcb.com/help/article/96-PCB-Panelization) is the relevant support page from JLC...yea unfortunately looks like they're not fans of two boards in a single gerber

PR #221, opened 2023-02-19
Add sensing carrier board for MKVI rear sensing box

jack-greenberg:
This board will connect to the primary Deutsch DT connector that forms the seal on the rear sensing box. It has a buck converter to step battery voltage down to +5V.

Note that the wheel-speed sensor is rated for a minimum of +6V, but I recently tested it with a +5V supply and it worked fine. Need to continue investigation. One option is to power it off of the battery and then use an open-drain circuit to send the signal to the sensing controller. TBD.

Still need to add silkscreen. Final layout pending completion of the sensing controller in #218.


richardli03 (changes requested):
Hey jack! Thanks so much for doing the carrier board. A couple of questions/potential problems? 

- The DRC is telling me 3 diff kinds of errors: Via diameter, Drill out of range, and a courtyard overlap. It seems like the vias and drills (though I'm not sure what those are) are barely too small for the required minimum diameter. The courtyards overlapping thing might just be the way the connector slots on so I'm pretty sure that's not a problem (?)
- I think these aren't quite the updated sensing controller board dimensions (which shouldn't change anymore thanks for your patience)

Those were the things I'd point out. Thanks again!

jack-greenberg:
Thanks for the review! All of the DRC errors can be safely ignored. The via sizes are within the manufacturer's capabilities and you were correct about the courtyard overlap. Just need to fix a couple things with the sensing board outline and fix some merge conflicts and then I'll be ready to merge.

amit-kumarh (approved):
LGTM!

PR #223, opened 2023-03-04
Add service section motherboard schematic and layout

vrhinehart:
I have completed a schematic for the service section motherboard, and I would appreciate any feedback or fixes people may have. [@hv](https://olinformula.slack.com/admin/user_groups) if you designed a module, I would really appreciate if you could check that I hooked it up correctly.

There are a few open questions still:

I’m thinking of using 3-pin connectors for the 2-pin HV connections so they cannot be accidentally confused with the 2-pin LV connectors.

The IMD grounding question above is still an open question.

I’m still not 100% clear on whether TSAL_FLAG needs to be connected to anything. I currently think no. I have also added a fuse to the TSAL/AIL module’s 12V input because the module itself does not have one. Please tell me if you think this is extraneous given that the board has its own 12V fuse.
[Motherboard PDF.pdf](https://github.com/olin-electric-motorsports/olin-electric-motorsports/files/10887293/Motherboard.PDF.pdf)


vrhinehart:
I have pushed a layout. I believe it's basically complete, save for potential on-board AIL. I would appreciate a review of the layout, as well as any schematic issues that need to be resolved before the board can be finalized. 
Thanks!

bubbleebubbles (changes requested):
## Schematic 
* For TSAL_FLAG, I would add a small test point. The TSAL module should be tested fully and working before added to the motherboard, but i can imagine a case where something goes wrong once everything’s been assembled (ie. at comp) and we need quick reference points. 
* Looks like the IMD has 3 pins that need to be grounded and since we don’t ground the bolts inside the accumulator, we’ll have to ground all three through a PCB. 
d7573725869cc32081694340630a16ecec503cafacbabbf3d92691b0f584f233dfb020e412208ef8973e77d3c2e4a970d58a41886f01de44689a87baa27d7089358d02ca69cc41167b6dcb19eb5c5d397f2f5c411e5bf933cab7fe5b14d5641515bd3475

* label which pins on the IMD connector correspond to which pin on the IMD Motherboard connector for easy reference when harnessing. 
* Add a pulldown on the drivetrain_cooling_lsd pin to prevent a floating voltage at the gate (Q1) 
* It would be useful to add an LED for the relay for debugging purposes -- to know whether the relay closed or not without having to plug/unplug things. 
* Everything else looks good, great job!

## Layout 
* Overall, layout looks good. 
* Please add distances between LV and HV in silkscreen so that it’s easy to show at comp that we’re following HV-LV separation rules 12.7mm is the required distance I believe. The box outlines are great just add a couple labels around with the measure tool. 

Great job on both :)

vrhinehart:
@bubbleebubbles The two chassis ground pins are going to be grounded to the chassis through the accumulator mounting tabs. Grounding them through the PCB would mean the IMD could not detect loss of chassis ground. The current plan is to use copper tape to bring that connection down the side of the accumulator enclosure to the mounting points. 

I added a 10K pulldown to the DRIVETRAIN_COOLING_LSD_TRIG pin, assuming that's what you meant.

The relay is a shutdown sensing node, so it will be sensed though CAN or via the AIR control leds a few inches away. There is a similar relay on the IMD latch module. Neither will be visible when the whole thing is closed and powered anyway. Do you still think I should add an LED on the mobo?

Will do on anything I didn't mention.

bubbleebubbles:
> @bubbleebubbles The two chassis ground pins are going to be grounded to the chassis through the accumulator mounting tabs. Grounding them through the PCB would mean the IMD could not detect loss of chassis ground. The current plan is to use copper tape to bring that connection down the side of the accumulator enclosure to the mounting points.

makes sense, ye it slipped my mind that the IMD needs to detect loss of chassis ground. Just wanted to make sure yall had a plan for that. maybe take a note of it in the schematic or in your documentation (apologies if you already have and i missed it)  

> The relay is a shutdown sensing node ...

ye, then no point in adding it. Thanks for clearing that up for me. 

everything else looks good.

jack-greenberg:
All looks good to me. Just make sure to add in the AIL LEDs to the top-side of the board and add a connector that accepts wires from the TSAL/AIL module.

PR #225, opened 2023-03-09
Add MkVI BSPD Firmware

Eemac:


vehicle/mkvi/software/brakes/bspd.h
```diff
@@ -0,0 +1,38 @@
...
+//Input for shutdown sense line
+//gpio_t BSPD_SHUTDOWN_SENSE = PC0;
+
+//////////////////////// GPIO - ANALOG /////////////////////////
+adc_pin_e BRAKE_PRESSURE_SENSE = ADC7;
+
+//Timer config
+void timer0_callback(void);
+
+timer_cfg_s timer0_cfg = {
```
amit-kumarh, line 28: Can you add a comment here saying what the timer configuration does? (ie run the 100hz clock for the main loop)
It's something isn't immediately obvious to a reader who isn't already acquainted with the common timer configs we use on the team :)

vehicle/mkvi/software/brakes/bspd.c
```diff
@@ -0,0 +1,118 @@
...
+
+int main(void) {
+    /////////////////////////////// BSPD STARTUP ///////////////////////////////
+    //Not sure what this does
+    sei();
+
+    //Auto-generated
+    can_init_bspd();
+
+    //Set up the internal ADC on the Arduino - configure it to read analog brake pressure values
```
amit-kumarh, line 77: nit: this is not an Arduino - you designed this board!

vehicle/mkvi/software/brakes/bspd.c
```diff
@@ -0,0 +1,118 @@
...
+    gpio_set_mode(BSPD_LL, INPUT);
+    pio_set_mode(BRAKELIGHT_LL, INPUT);
+
+    //Attach Pins to interrupt handler (assuming on rising/falling edge)
+    //These are all PB registers which are mapped to the pcint0_callback function
+    gpio_enable_interrupt(BRAKELIGHT_LL);
+    gpio_enable_interrupt(MOTOR_CURRENT_SENSE);
+    gpio_enable_interrupt(BSPD_LL);
+
+    pcint1_callback();
```
amit-kumarh, line 102: This should be `pcint0_callback`

amit-kumarh (changes requested):
Overall, super awesome job!!! Mostly just left some minor comments, but I wouldn't be shocked if we find a few more bugs when we try to compile. Let me know if you have any questions

General Notes:
- Can you please change the name of the PR to something a little more descriptive, like "Add MKVI Brakes Firmware"?

EDIT: One thing  I missed! In line 13 of your YAML - the ss_bspd value is a boolean, and thus only takes 1 bit (like the other booleans at the bottom of the file), not the 8 bits that are currently allocated

vehicle/mkvi/software/brakes/bspd.h
```diff
@@ -0,0 +1,38 @@
...
+
+// Monitor Pins connected to the logic-level (LL) side of the LSDs
+gpio_t BSPD_LL = PB3;
+gpio_t BRAKELIGHT_LL = PB4;
+
+// Digital sense pin for 5kW motor "on" state
+gpio_t MOTOR_CURRENT_SENSE = PB5;
+
+// Input for shutdown sense line
+// gpio_t BSPD_SHUTDOWN_SENSE = PC0;
```
amit-kumarh, line 20: Ah, OK, so I've just caught something now that should have been caught in board review - what we're normally doing with the shutdown sense circuits is we literally want to feed a branch off of the shutdown circuit through the SS circuit to ensure there are no failure points measuring at some logic level point and the actual shutdown circuit. The shutdown sensing on your board takes the logic level as the input, which doesn't really make that distinction clear (and makes that circuit a bit redundant).

It's not the biggest of deals - as long as the relay works, which there isn't any reason to believe it won't, the values will be equivalent. Just wanted to make note of it here. May also be worth adding a comment as to why we're not using this value since that will be the convention elsewhere on the car for the sake of future readers.

vehicle/mkvi/software/brakes/bspd.c
```diff
@@ -0,0 +1,118 @@
...
+
+// Code to run when a digital sense pin changes state (High->Low or Low->High)
+volatile bool updateLEDsTrigger = false;
+void pcint0_callback(void) {
+    // Update CAN struct with new board logic values
+    bspd.brake_gate = !!gpio_get_pin(BRAKELIGHT_LL);
+    bspd.bspd_5kw = !!gpio_get_pin(MOTOR_CURRENT_SENSE);
+    bspd.ss_bspd = !gpio_get_pin(BSPD_LL);
+
+    updateLEDsTrigger = true;
```
amit-kumarh, line 37: can we `snake_case` to be consistent with everywhere else?

vehicle/mkvi/software/brakes/bspd.h
```diff
@@ -0,0 +1,38 @@
...
+
+//////////////////////// GPIO - ANALOG /////////////////////////
```
amit-kumarh, line 22: nit: the GPIO usually refers exclusively to digital pins, so analog inputs aren't part of that

vehicle/mkvi/software/brakes/bspd.c
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Check whether an LED needs updating, and if so, change its state
+void updateLEDs(void) {
```
amit-kumarh, line 41: `snake_case` here as well

amit-kumarh (changes requested):
Awesome job!!! very minor nits here mainly - will loop in Jack on this for a final review. Once we get the board tested and make fixes if necessary there, we can approve and merge. 

Thanks for your great great work on this!

jack-greenberg (changes requested):

amit-kumarh (approved):
LGTM (insert obama meme)

Just removed all the HitL commits to get this ready to merge

PR #227, opened 2023-03-22
Add MKVI Throttle Firmware

eys123:
Reads throttle potentiometers and checks for deviation, brake, and out of range implausibilities. If there are no implausibilities, returns each throttle pot value.
Changes from last year: there are no debugging LEDs on my board this year, so I added CAN debugging messages to throttle.yml instead. We also only have the inertia switch pin on the throttle board this year.
Right now the compile is failing because the brakes .yml file isn't in the MKVI .dbc yet, but my code will compile outside of that.


vehicle/mkvi/software/throttle/throttle.yml
```diff
@@ -0,0 +1,83 @@
...
+          scale: 100/255
+      - name: ss_is
+        slice: 41 + 1
+        unit:
+          type: bool
+          values:
+            t: CLOSED
+            f: OPEN
+  - name: throttle_debug
+    id: 0x5C0
```
ayushchakra, line 46: I think this ID should be 0x6C0 (based on https://docs.google.com/spreadsheets/d/12KenzPC5moFQULhOhoBm-pcmC4ligmMhP_M-TPqN-CI/edit#gid=1599758060)

vehicle/mkvi/software/throttle/throttle.c
```diff
@@ -0,0 +1,322 @@
...
+            if (can_poll_receive_dashboard() == 0) {
+                // check RTD status
+                can_receive_dashboard();
+                bool ready_to_drive = dashboard.ready_to_drive;
+                if (ready_to_drive) {
+                    m192_command_message.inverter_enable = true;
+                } else {
+                    m192_command_message.inverter_enable = false;
+                }
+            }
```
ayushchakra, line 269: nit: This could be simplified to just assigning the ready to drive status to the inverter_enable status instead of creating a local variable to store it

vehicle/mkvi/software/throttle/throttle.c
```diff
@@ -0,0 +1,322 @@
...
+        *pos_2 = 0;
+        throttle.throttle_state = THROTTLE_OUT_OF_RANGE;
+        throttle_debug.throttle_out_of_range = true;
+        implausibility = true;
+    } else {
+        // Will be reset if Ready to Drive is not on
+        throttle.throttle_state = THROTTLE_RUN;
+        throttle_debug.throttle_out_of_range = false;
+    }
+    return implausibility;
```
ayushchakra, line 124: With your out of range debug message, if the first position is out of range, but the second isn't, you're overriding the `throttle_debug.throttle_out_of_range` value, meaning that the CAN message won't say there's an out of range error even if there is. I suggest splitting up the out of range signal into two separate values, one for each pot, in the CAN message.

ayushchakra: Same for `throttle.throttle_state`

vehicle/mkvi/software/throttle/throttle.c
```diff
@@ -0,0 +1,322 @@
...
+                SET_TORQUE_REQUEST(0);
+                continue;
+            }
+
+            if (oor_implausibility || deviation_implausibility) {
+                throttle_state.implausibility_fault_counter++;
+                if (throttle_state.implausibility_fault_counter
+                    >= IMPLAUSIBILITY_TIME_LIMIT) {
+                    // if timer reaches IMPLAUSIBILITY_TIMEOUT
+                    // cut power and reset timer
```
ayushchakra, line 292: Is this cutting power, or just requesting 0 torque from the motor controller?

ayushchakra (commented):
Overall, this looks great, awesome work! I just left a few comments. Let me know if you have any questions or if anything I said seems wrong. Also highly suggest writing unit tests for your code, it's the best way to review and test your code without having the board at hand!

vehicle/mkvi/software/throttle/throttle.c
```diff
@@ -0,0 +1,322 @@
...
+static bool check_brake(int16_t pos_min) {
+    /*
+        Check if brakes are on while pedal travel >= 25%, if so return
+        implausibility If pedal travel becomes <5%, clear implausibility.
+        Args:
+            pos_min: int16_t smaller sensor value of the 2 pots
+        Returns:
+            true if implausibility, false if not
+    */
+    static bool brake_implausibility_occurred = false;
```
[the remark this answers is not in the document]
amit-kumarh: I believe in this case, the use of static isn't to access the variable outside the function, but to ensure the value stays persistent between function calls, so we can know if we were previously in an implausibility state the last time we ran this function.

In that scenario, making a static variable would maintain it at the lowest possible scope (function scope) as opposed to making a global variable.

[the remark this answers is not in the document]
ayushchakra: Ah that makes sense, thought it was being used elsewhere. Thanks for clarifying.

vehicle/mkvi/software/throttle/throttle_config.h
```diff
@@ -0,0 +1,103 @@
...
+gpio_t SS_IS = PB6;
+
+typedef struct {
+    // represents a throttle poteniometer
+    adc_pin_e adc_pin;
+    int16_t throttle_min; // Minimum position (0%) in raw ADC
+    int16_t throttle_max; // Maximum position (100%) in raw ADC
+} throttle_potentiometer_s;
+
+const throttle_potentiometer_s throttle_1 = {
```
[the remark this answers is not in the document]
eys123: I have a couple of other variables indexed starting from 1, is it alright if I just keep this?

jack-greenberg:
Throttle variabl names should use R and L, not 1 and 2. I messed up last year and it was super confusing, so i recommend changing to R and L

vehicle/mkvi/software/throttle/throttle.yml
```diff
@@ -0,0 +1,91 @@
...
+            - THROTTLE_IDLE
+            - THROTTLE_RUN
+            - THROTTLE_1_OUT_OF_RANGE
+            - THROTTLE_2_OUT_OF_RANGE
+            - THROTTLE_POSITION_IMPLAUSIBILITY
+            - THROTTLE_BRAKE_PRESSED
+      - name: throttle_1_pos
+        slice: 8 + 16
+        unit:
+          type: int16_t
```
ayushchakra, line 28: Shouldn't this be a int8_t since the slice is only 8 bits?

vehicle/mkvi/software/throttle/throttle.c
```diff
@@ -0,0 +1,318 @@
...
+            throttle_debug.throttle_brake_implaus = true;
+            return true;
+        } else {
+            // brake is pressed, pedal travel < 25%
+            // no implausibility
+            throttle_debug.throttle_brake_implaus = false;
+            return false;
+        }
+
+        if (brake_implausibility_occurred) {
```
ayushchakra, line 174: Doesn't this if statement never get executed since the previous block is guaranteed to return something?

vehicle/mkvi/software/throttle/throttle.yml
amit-kumarh: If memory serves me, the first value is the starting bit and the second is the slice length, making this a 16 bit slice.

ayushchakra: Oh you're right, misread that.

amit-kumarh (approved):
Sep 2023 status update: code was tested SP 23 and should be pretty much good to go. It's not building without a dashboard.yml.

TODO: Calibrate the APPS values for MKVI

It won't build without a dashboard.yml - if we calibrate the pedals, feel free to ignore and merge (assuming it builds with your local, which you can test just be adding the MKV dashboard yaml to the `//vehicle/mkvi/BUILD` file)

amit-kumarh (approved):
merging in for wheels spinning push

PR #230, opened 2023-03-28
Fix syntax errors in symbol library

drewnotdrew:
The recent PR #215 had issues merge conflicts with the symbol library that were thought to have been resolved with a rebase by @elviswolcott. However, there were still syntax errors that cause various issues with the symbol library.

Attempting to load the current formula `symbol` library does not work because of syntax errors: 
 broken

Attempting to add new parts to the symbol library removes all symbols past the syntax error.

This PR fixes the syntax errors and removes a problematic symbol that is also available in one of the KiCad libraries.


elviswolcott (approved):
Thanks for figuring this out!

PR #234, opened 2023-04-08
sensing firmware draft

richardli03:
Just looking to get feedback to make sure I'm not making any critical mistakes moving forward

### DIFF
--- a/vehicle/mkvi/software/sensing/BUILD
+++ b/vehicle/mkvi/software/sensing/BUILD
@@ -1,28 +1,13 @@
 load("//bazel/tools:defs.bzl", "can_api_files")
-load("//bazel:defs.bzl", "cc_firmware")
+# load("//bazel:defs.bzl", "cc_firmware")
+load (":build.bzl", "sensing_firmware")
 
 exports_files([
     "sensing.yml",
 ])
 
-cc_firmware(
-    name = "sensing",
-    srcs = [
-        "sensing.c",
-        "sensing.h",
-    ],
-    # btldr = "//projects/btldr:brakelight_bspd_btldr",
-    target_compatible_with = [
-        "//bazel/constraints:atmega16m1",
-    ],
-    visibility = ["//visibility:public"],
-    deps = [
-        ":can_api",
-        "//libs/gpio",
-        "//libs/timer",
-        # "//projects/btldr:btldr_lib",
-    ],
-)
+sensing_firmware("front")
+sensing_firmware("rear")
 
 can_api_files(
     name = "can_api",
--- a/vehicle/mkvi/software/sensing/build.bzl
+++ b/vehicle/mkvi/software/sensing/build.bzl
@@ -1,0 +1,30 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+def sensing_firmware(location):
+    defines = {}
+
+    if location == "front":
+        defines["SENSING_LOCATION"] = 0
+
+    if location == "rear":
+        defines["SENSING_LOCATION"] = 1
+
+    cc_firmware(
+        name = "sensing_{}".format(location),
+        srcs = [
+            "sensing.c",
+            "sensing.h",
+        ],
+        defines = defines,
+        target_compatible_with = [
+            "//bazel/constraints:atmega16m1",
+        ],
+        visibility = ["//visibility:public"],
+        deps = [
+            ":can_api",
+            "//libs/adc",
+            "//libs/gpio",
+            "//libs/timer",
+            "//projects/btldr:btldr_lib",
+        ],
+    )


vehicle/mkvi/software/sensing/sensing.yml
```diff
@@ -0,0 +1,19 @@
+name: brakelight_bspd
```
ayushchakra, line 1: Update the board name to sensing

vehicle/mkvi/software/sensing/sensing.yml
```diff
@@ -0,0 +1,19 @@
...
+# subscribe:
+#   - name: air_control_critical
+#     mob: 1
+publish:
+  - name: sensing
+    id: 0xb
```
ayushchakra, line 7: Make sure you update the CAN ID based on our address sheet (https://docs.google.com/spreadsheets/d/12KenzPC5moFQULhOhoBm-pcmC4ligmMhP_M-TPqN-CI/edit#gid=1599758060). Not sure how up to date the Sensing section is on the document based on this year's architecture.

vehicle/mkvi/software/sensing/BUILD
```diff
@@ -0,0 +1,32 @@
...
+    "sensing.yml",
+])
+
+cc_firmware(
+    name = "sensing",
+    srcs = [
+        "sensing.c",
+        "sensing.h",
+    ],
+    btldr = "//projects/btldr:brakelight_bspd_btldr",
```
ayushchakra, line 14: Maybe a question for @jack-greenberg but is there a separate btldr config for the sensing board?

vehicle/mkvi/software/sensing/BUILD
```diff
@@ -0,0 +1,32 @@
...
+        "//libs/adc",
+        "//libs/gpio",
+        "//libs/timer",
+        "//projects/btldr:btldr_lib",
+    ],
+)
+
+can_api_files(
+    name = "can_api",
+    dbc = "//vehicle/mkv:mkv.dbc",
```
ayushchakra, line 30: MKVI DBC not mkv

ayushchakra: Also side note but make sure u add the sensing yaml to the mkvi dbc in `vehicle/mkvi/BUILD`

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,73 @@
+/*
+Header:
+    Code for the Sensing Controller Board located in the nostril motherboard and
```
ayushchakra, line 3: I'm probably very behind on the current sensing architecture, but is the sensing firmware specific to front and rear boards? If so, probably should create different directories for both

vehicle/mkvi/software/sensing/sensing.h
```diff
@@ -0,0 +1,37 @@
...
+/*
+ * Timer config
+ */
+void timer1_isr(void);
+
+/*
+ * 1s timer (1kHz)
+ *
+ * F_CPU / prescaler / output_compare_match
+ * 4000000 / 256 / 15625 = 1
```
ayushchakra, line 22: Comment math doesn't match the prescaler you picked

vehicle/mkvi/software/sensing/sensing.h
```diff
@@ -0,0 +1,37 @@
...
+#include "libs/gpio/pin_defs.h"
+#include "libs/timer/api.h"
+
+/* ATMega 16m1 Pins */
+gpio_t DEBUG_LED = PB0;
+gpio_t WHEEL_SPEED_LEFT = PB2;
+gpio_t WHEEL_SPEED_RIGHT = PD6;
+
+enum { counter_left = 0 };
+enum { counter_right = 0 };
```
ayushchakra, line 12: Are these supposed to be within one enum?

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,73 @@
...
+#include "utils/timer.h"
+#include <avr/interrupt.h>
+#include <avr/io.h>
+#include <stdio.h>
+#include <string.h>
+#include <util/delay.h>
+
+volatile uint8_t gTimerFlag = 0x01;
+uint8_t gStatusMessage[8] = {};
+volatile uint8_t can_recv_msg[8] = {};
```
ayushchakra, line 22: Doesn't the can api handle allocating mem for CAN messages?

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,73 @@
...
+    gpio_set_mode(WHEEL_SPEED_RIGHT, INPUT);
+
+    sei(); // Initializes interrupts
+    gpio_enable_interrupt(WS1);
+    gpio_enable_interrupt(WS2);
+
+    timer_init(&timer1_cfg);
+
+    pcint0_callback();
+    pcint2_callback();
```
ayushchakra, line 58: Little confused by this. Don't these functions need arguments, also why do interrupts need to be called in your main function, shouldn't they just be called based on the gpio channel?

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,73 @@
...
+uint8_t gStatusMessage[8] = {};
+volatile uint8_t can_recv_msg[8] = {};
+
+void timer1_isr(void) {
+    send_can = true;
+};
+
+int pcint0_callback(int counter_left) {
+    // Left wheel speed sensor is in pin PB2 -> PORTB
+    sensing.wsl = !gpio_get_pin(WHEEL_SPEED_LEFT);
```
ayushchakra, line 30: nit: You use `ws1` for the left sensor, but `wsr` for the right. Make sure you keep your naming convention consistent, l and r make sense to me.

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,73 @@
...
+#include "can_api.h"
+#include "log_uart.h"
+#include "utils/timer.h"
+#include <avr/interrupt.h>
+#include <avr/io.h>
+#include <stdio.h>
+#include <string.h>
+#include <util/delay.h>
+
+volatile uint8_t gTimerFlag = 0x01;
```
ayushchakra, line 20: What are `gTimerFlag` and `gStatusMessages`? Might be nice to include a comment about their function

vehicle/mkvi/software/sensing/sensing.h
ayushchakra: Oh I kept reading, seems like these are just counters? Is there a reason why you didn't just have these as 8 or 16 bit integers?

vehicle/mkvi/software/sensing/sensing.c
ayushchakra: Maybe a better structure for this is you had two volatile uint8s that increment on every interrupt call?

vehicle/mkvi/software/sensing/sensing.yml
```diff
@@ -0,0 +1,19 @@
+name: brakelight_bspd
+# subscribe:
+#   - name: air_control_critical
+#     mob: 1
```
ayushchakra, line 4: Does the sensing board need to listen to any CAN messages?

richardli03: I don't believe so! It's just reporting what it gets from wheel speed sensors

ayushchakra: What's consuming this data? Dashboard to send over telemetry?

jack-greenberg: Data is just being logged for post-drive analysis.

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: Not sure about the context here specifically, but I have used this pattern in the past. `pcintN_callback()` are functions that are called by the interrupts for PORTx. The nice thing about having them as standalone functions is that we can also use them as regular functions. Like in the AIR control code, the interrupts are triggered anytime the state of the pins change, but we also call the functions directly to capture the state of the pins during initialization.

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: They will be the same, but we need to figure out some way to differentiate them so they know which CAN messages to send out. Probably EEPROM makes the most sense. @richardli03 I can work on that once your code is good to go.

vehicle/mkvi/software/sensing/BUILD
jack-greenberg: There will be! @richardli03 please comment out everything relating to the bootloader. I will create a PR adding the bootloader to all MKVI code.

vehicle/mkvi/software/sensing/sensing.yml
richardli03: @jack-greenberg could I just assign myself on the spreadsheet for this?

vehicle/mkvi/software/sensing/BUILD
richardli03: got it! the path i'm using is in `vehicle/mkvi/build` is `"//vehicle/common/sensing:sensing.dbc"` just based on pattern matching -- not sure if that's correct

vehicle/mkvi/software/sensing/sensing.h
richardli03: There was something with the interrupts that made @amit-kumarh suggest just making them global variables. It makes the syntax easier (?)

vehicle/mkvi/software/sensing/sensing.c
richardli03: Yeah, each of these callbacks are triggered whenever the wheel speed sensor sees metal, and the interrupt increments the L + R counters (depending on which pin). I _believe_ this is the correct way to do this?

vehicle/mkvi/software/sensing/sensing.c
richardli03: deleted from prev comment!

vehicle/mkvi/software/sensing/sensing.h
amit-kumarh: They can be global variables and `uint8_t`s or something like that would be what I would say. I would also for readability sake maybe define this in the .c file, since it isn't really configuration? up to you

edit: also potentially see my other comment about getting rid of these

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,69 @@
...
+#include <avr/io.h>
+#include <stdio.h>
+#include <string.h>
+#include <util/delay.h>
+
+void timer1_isr(void) {
+    send_can = true;
+};
+
+int pcint0_callback(int counter_left) {
```
amit-kumarh, line 24: If your two counters are global variables, you don't need to take them in as function parameters

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,69 @@
...
+
+    pcint0_callback();
+    pcint2_callback();
+    timer1_isr();
+    while (true) {
+        // Check if the timer is at 1 second (arbitrary)
+        // If timer is at 1 second
+        if (send_can) {
+            // Log how many times we saw metal
+            sensing.left_wheel_counter = counter_left;
```
amit-kumarh, line 61: Also, something I just thought of that would reduce our memory footprint and might be a little easier - instead of incrementing separate counters, and then assigning those values into our CAN struct, we can just increment the struct fields in our GPIO interrupt. So they might look something like:

```
void pcint2_callback(int counter_right) {
    // Right wheel speed sensor is in pin PB2 -> PORTB
    sensing.wsr = !gpio_get_pin(WHEEL_SPEED_RIGHT);
    if (sensing.wsr) {
        sensing.right_wheel_counter += 1;
    }
}
```

and then in this block you can just send CAN and reset your counters to 0

vehicle/mkvi/software/sensing/sensing.h
amit-kumarh: also, make sure you change your CTC value in the timer config as well!

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,69 @@
...
+    Code for the Sensing Controller Board located in the nostril motherboard and
+the rear sensing box Author:
+    @author Richard Li
+    @date   03/25/2023
+*/
+
+/*----- Includes -----*/
+#include "sensing.h"
+#include "can_api.h"
+#include "log_uart.h"
```
amit-kumarh, line 12: Is this import used anywhere?

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,69 @@
...
+
+int pcint0_callback(int counter_left) {
+    // Left wheel speed sensor is in pin PB2 -> PORTB
+    sensing.wsl = !gpio_get_pin(WHEEL_SPEED_LEFT);
+    if (sensing.wsl) {
+        counter_left += 1;
+    }
+}
+void pcint2_callback(int counter_right) {
+    // Right wheel speed sensor is in pin PB2 -> PORTB
```
amit-kumarh, line 32: this comment should say PORTD?

vehicle/mkvi/software/sensing/sensing.c
```diff
@@ -0,0 +1,69 @@
...
+    sensing.wsr = !gpio_get_pin(WHEEL_SPEED_RIGHT);
+    if (sensing.wsr) {
+        counter_right += 1;
+    }
+}
+
+int main(void) {
+    can_init_sensing();
+    adc_init();
+    CAN_init(CAN_ENABLED);
```
amit-kumarh, line 42: What is this line doing? Don't think we need it for CAN API

amit-kumarh (changes requested):

vehicle/mkvi/software/sensing/sensing.h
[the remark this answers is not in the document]
richardli03: done!

vehicle/mkvi/software/sensing/sensing.c
richardli03: probably just going to delete them entirely and increment within the struct like you mentioned!

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: Additionally, those functions cannot take arguments because they are also used in the GPIO library where they are called without arguments.

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: No ADCs for now.

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: This is a relic from old code and should be removed.

vehicle/mkvi/software/sensing/sensing.c
jack-greenberg: @richardli03 In the interrupts (in these functions) you might need to check to see if the pin has gone high/low before incrementing. We only want to increment each time it goes in _one direction_, but I'm not sure if it's when it goes high or low. Basically we want to increment every time we get to a hole (no metal) and I'm not sure if that's high or low from the WSS. We can test it once we have boards done.

vehicle/mkvi/software/sensing/sensing.c
amit-kumarh: I was referring to `CAN_init(CAN_ENABLED);`, but yeah nice catch on the ADCs

PR #236, opened 2023-04-10
EEPROM configuration for sensing board location

jack-greenberg:
This PR adds an EEPROM configuration that tells the sensing controller which location it is in (front vs rear) so that it knows which CAN messages to send out.

### DIFF
--- a/vehicle/mkvi/software/sensing/BUILD
+++ b/vehicle/mkvi/software/sensing/BUILD
@@ -1,29 +1,13 @@
 load("//bazel/tools:defs.bzl", "can_api_files")
-load("//bazel:defs.bzl", "cc_firmware")
+# load("//bazel:defs.bzl", "cc_firmware")
+load (":build.bzl", "sensing_firmware")
 
 exports_files([
     "sensing.yml",
 ])
 
-cc_firmware(
-    name = "sensing",
-    srcs = [
-        "sensing.c",
-        "sensing.h",
-    ],
-    btldr = "//projects/btldr:brakelight_bspd_btldr",
-    target_compatible_with = [
-        "//bazel/constraints:atmega16m1",
-    ],
-    visibility = ["//visibility:public"],
-    deps = [
-        ":can_api",
-        "//libs/adc",
-        "//libs/gpio",
-        "//libs/timer",
-        "//projects/btldr:btldr_lib",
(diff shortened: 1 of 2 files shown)


jack-greenberg:
With the changes I've added, in the sensing firmware, you'll need to add the following:

```c
// sensing.c
#include <avr/eeprom.h>

enum sensing_location {
    FRONT = 0,
    REAR = 1
} location __attribute__((section(".eeprom"))) = SENSING_LOCATION;
```

```c
can_frame_t sensing_msg = {
    .mob = 0,
    .dlc = 2,
};

int main(void) {
    eeprom_busy_wait();
    enum sensing_location location = eeprom_read_dword(&location);
    _delay_ms(2);

    switch (location) {
        case FRONT: {
            sensing_msg.id = 0x403;
        } break;
        case REAR: {
            sensing_msg.id = 0x404;
        } break;
    }

    ...
}
```

Because we need to be able to dynamically determine which CAN message you are sending, you'll need to use the CAN library instead of the CAN API.

richardli03:
> With the changes I've added, in the sensing firmware, you'll need to add the following:
> 
> ```c
> // sensing.c
> #include <avr/eeprom.h>
> 
> enum sensing_location {
>     FRONT = 0,
>     REAR = 1
> } location __attribute__((section(".eeprom"))) = SENSING_LOCATION;
> ```
> 
> ```c
> can_frame_t sensing_msg = {
>     .mob = 0,
>     .dlc = 2,
> };
> 
> int main(void) {
>     eeprom_busy_wait();
>     enum sensing_location location = eeprom_read_dword(&location);
>     _delay_ms(2);
> 
>     switch (location) {
>         case FRONT: {
>             sensing_msg.id = 0x403;
>         } break;
>         case REAR: {
>             sensing_msg.id = 0x404;
>         } break;
>     }
> 
>     ...
> }
> ```
> 
> Because we need to be able to dynamically determine which CAN message you are sending, you'll need to use the CAN library instead of the CAN API.

Does this mean I can't reference the CAN API at all within my code + have to change my current calls to API?

richardli03 (approved):
definitely do not feel confident at all implementing these changes. I don't want to step too much on your bandwidth, but could I ask you to flesh some of this out with me more at some point?

PR #238, opened 2023-04-22
Update power board with more vias

jack-greenberg:


elviswolcott (approved):
Van and I took a look together and this looks great!

PR #240, opened 2023-05-01
Throttle HITL Tests

ayushchakra:
Remaining Tasks:

- [ ] Rebase onto HITL Software Refactor
- [ ] Update `set_throttle_voltages` to incorporate Throttle Transfer Equations (Same voltage does not mean same position - the current tests will not pass!)
- [ ] Add torque request checks across the board
- [ ] Use "L" and "R" instead of 1 and 2 for throttle names
- [ ] Overall code cleanliness (too much duplicate code, abstract out assertions?)

### DIFF
--- a/vehicle/mkvi/software/air_control/test_air_control.py
+++ b/vehicle/mkvi/software/air_control/test_air_control.py
@@ -1,0 +1,11 @@
+# Base/extended python imports
+import pytest
+import os
+import time
+from configparser import ConfigParser
+from can import Message
+
+# Project Imports
+from projects.hitl.software.hitl.cancontroller import CANController
+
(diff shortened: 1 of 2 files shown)


vehicle/mkvi/software/air_control/test_air_control.py
```diff
@@ -0,0 +1,11 @@
+# Base/extended python imports
```
ayushchakra, line 1: @richardli03 Do we still need this file?

jack-greenberg: This will get updated as part of #231. You can remove from this PR.

PR #242, opened 2023-06-09
IMU Rev 2.1

drewnotdrew:
Minor version bump to resolve issues with LDO part sourcing.

Changes:
- Replaced 5V and 1.8V LDOs with more common pinout versions. Should the currently specced parts go out of stock, they can be easily replaced
- Updated front and back silkscreens for aesthetics


mkazc (approved):
After visual checks and DRC, this looks all good to go! Awesome work!

PR #244, opened 2023-09-09
Rev 2 of MKVI Nostril Motherboard

amit-kumarh:
- Make Throttle 1+ and 2+ lines independent. This allows us to use the inline resistors on the throttle board instead having to find a spot to place other ones (on r1 I accidentally shorted both those nets to 5V, bypassing those resistors)
- Move the fuse further away from the ampseal (was difficult to hand solder)
- update to latest symbols/footprints


vrhinehart (approved):
I can’t find anything wrong, it looks like what was proposed matches what was done.

PR #245, opened 2023-09-10
Firmware Session 1 Homework

drewnotdrew:
Added a `main.c` to test the power_distribution library. Tested and working with a board with 3 LEDs.

Questions:
- Is there a better way to expose the devices outside of the power distribution library than using `extern` in `api.h`?
- Is there a better place to put `gpio_set_mode` for each device instead of an explicit `init_devices` function?
- Is there a better way to power all devices and power off all devices other than making repeated calls to set and clear the pins?

### DIFF
--- a/tutorials/power_distribution/api.h
+++ b/tutorials/power_distribution/api.h
@@ -1,0 +1,14 @@
+#pragma once
+
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+extern gpio_t DASHBOARD_POWER;
+extern gpio_t SERVICE_SECTION_POWER;
+extern gpio_t THROTTLE_POWER;
+
+void init_devices();
+void power_device();
+void power_all();
+void power_off_device();
+void power_off_all();
--- a/tutorials/power_distribution/api.h
+++ b/tutorials/power_distribution/api.h
@@ -1,14 +1,32 @@
 #pragma once
 
 #include "libs/gpio/api.h"
-#include "libs/gpio/pin_defs.h"
 
 extern gpio_t DASHBOARD_POWER;
 extern gpio_t SERVICE_SECTION_POWER;
 extern gpio_t THROTTLE_POWER;
 
-void init_devices();
-void power_device();
-void power_all();
-void power_off_device();
-void power_off_all();
+/*
+ * Initialize power distribution device pins
+ */
+void init_devices(void);
+
+/*
+ * Power a device
+ */
+void power_device(gpio_t device);
+
+/*
+ * Power all devices
+ */
+void power_all(void);
+
+/*
+ * Power off a device
+ */
+void power_off_device(gpio_t device);
+
+/*
+ * Power off all devices
+ */
+void power_off_all(void);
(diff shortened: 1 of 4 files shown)


tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,14 @@
...
+
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+extern gpio_t DASHBOARD_POWER;
+extern gpio_t SERVICE_SECTION_POWER;
+extern gpio_t THROTTLE_POWER;
+
+void init_devices();
+void power_device();
```
amit-kumarh, line 11: make sure you're adding in arguments for the functions that have them! if there are no arguments, you can put void

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+
+extern gpio_t DASHBOARD_POWER;
+extern gpio_t SERVICE_SECTION_POWER;
+extern gpio_t THROTTLE_POWER;
+
+void init_devices();
```
amit-kumarh, line 10: also, in a real api.h file, I would be adding in docstrings so someone who wants to use this library can see how

tutorials/power_distribution/power_distribution.c
amit-kumarh, line 1: 👍

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
```
amit-kumarh, line 4: technically, I'm not sure if you need the pin_defs file since I don't think we're using them in this header

amit-kumarh:
To answer your questions:

1. Nah, that's typically how we do it. If I'm writing a firmware image, where my header file is only going to be used by my one main C file, then I can define variables and things in my header for that image (including GPIO variables, but also things like config structs, etc.); most of our small firmware images follow this pattern. But if I'm writing a library that will be used in multiple places, then you'd want to define them in a C file and make them available elsewhere with `extern`. The reason for this is the header file is what you would import into your other programs, and if you end up with a scenario where we import the same header into multiple places and we define some variables in it, then you'll run into compiler errors because things are being defined multiple times. For an example of this, check out the BMS firmware config library - it has a .h file and a .c file because we can't define variables in a .h that is used in multiple places.

2. sometimes for small images I'll just throw it right in the main function, so it's like 
```
int main() {
  // set up code including GPIO stuff
  
  for (;;) {
    // main loop
  }
}
```
but it can be neater to just have a `init_devices` or `hw_init` function or something like that.

3. For our libraries, nah, we don't have functions that affect multiple pins or anything like that. There is a `gpio_toggle_pin`, which just flips its current state. You can always look at the `api.h` files for our libraries to see what they can do

tutorials/power_distribution/api.h
```diff
@@ -0,0 +1,32 @@
...
+extern gpio_t SERVICE_SECTION_POWER;
+extern gpio_t THROTTLE_POWER;
+
+/*
+ * Initialize power distribution device pins
+ */
+void init_devices(void);
+
+/*
+ * Power a device
```
amit-kumarh, line 15: typically you'd also notate parameters for functions that require them but nbd here - you get the point

amit-kumarh (commented):
LGTM! not gonna approve since we don't wanna merge this. feel free to close the PR at your convenience

PR #247, opened 2023-10-05
Complete Super Blink

drewnotdrew:
Tested and working with a board with 2 leds.

Questions:
- Can all gpio pins use a hardware timer?
- When defining flags such as `send_can`  or `run_blink` , why might the `static` keyword be used if the scope is global? When would we go out of scope and still need either flag?

Still working through commenting the ADC library, and I'll re-request review and post more questions once I'm done.

### DIFF
--- a/tutorials/super_blink/BUILD
+++ b/tutorials/super_blink/BUILD
@@ -1,0 +1,14 @@
+load("//bazel:defs.bzl", "cc_firmware")
+
+cc_firmware(
+    name = "super_blink",
+    srcs = [
+        "blink_config.h",
+        "super_blink.c",
+    ],
+    target_compatible_with = ["//bazel/constraints:avr"],
+    deps = [
+        "//libs/gpio",
+        "//libs/timer",
+    ],
+)
--- a/tutorials/super_blink/blink_config.h
+++ b/tutorials/super_blink/blink_config.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include "libs/gpio/api.h"
+#include "libs/gpio/pin_defs.h"
+#include "libs/timer/api.h"
+
+/*
+ * Pin definitions
+ */
+gpio_t LED0 = PB5;
+gpio_t LED1 = PB6;
+
+/*
+ * Timer config
+ */
+
+void timer0_callback(void);
+
+timer_cfg_s timer0_cfg = {
+    .timer = TIMER0,
+    .timer0_mode = TIMER0_MODE_CTC,
+    .prescalar = CLKIO_DIV_1024,
+    .channel_a = {
+        .output_compare_match = 0x81,
+        .pin_behavior = DISCONNECTED,
+        .interrupt_enable = true,
+        .interrupt_callback = timer0_callback,
+    },
+};
--- a/tutorials/super_blink/research/adc.c
+++ b/tutorials/super_blink/research/adc.c
@@ -12,15 +12,29 @@ ISR(ADC_vect) {
     }
 }
 
-void adc_init(void) {
-    ADCSRA = (1 << ADEN) | (1 << ADPS2) | (1 << ADPS0);
-    ADCSRB &= (1 << AREFEN);
+/*
+ * Initialize the adc on the MCU
+ * ADCSRA = ADC register A containing configuration bits (ADC Control and Status Register A)
+ * ADCSRB = ADC register B containing more configuration bits
+ * ADEN = ADC enable bit
+ * ADPS2:0 = ADC prescaler select - division factor between system clock fq and input clock to ADC
+ * AREFEN = Analog reference pin enable
+ * ADMUX = Register for muxing to various ADC inputs
+ * REFS0 = REFSn: Selects reference for the ADC
+ */
 
-    ADMUX |= (1 << REFS0);
+// Initializes the ADC on the MCU
+void adc_init(void) {
+    ADCSRA = (1 << ADEN) | (1 << ADPS2) | (1 << ADPS0); // Sets the ADEN, ADPS2, and ADPS0 bits to 1. Enables the ADC, sets the prescaler to 32
+    ADCSRB &= (1 << AREFEN); // Sets the AREFEN bit to 1. 
+
+    ADMUX |= (1 << REFS0); // Bits: 01000000; ADC reference voltage is AV_CC
 }
 
-void adc_start_convert(adc_pin_e pin) {
-    ADMUX &= ~0x1F;
+
+// Starts ADC converting on a pin
+void adc_start_convert(adc_pin_e pin) { // Takes a type adc_pin_e named pin, maps back to an enum
+    ADMUX &= ~0x1F; // In binary: 11100000
     ADMUX |= (pin & 0x1F);
 
     ADCSRA |= (1 << ADSC);
--- a/tutorials/super_blink/research/adc.c
+++ b/tutorials/super_blink/research/adc.c
@@ -1,0 +1,54 @@
+#include "libs/adc/api.h"
+
+#include <avr/interrupt.h>
+#include <avr/io.h>
+#include <stdlib.h>
+
+void (*interrupt_callback)(void) = NULL;
+
+ISR(ADC_vect) {
+    if (interrupt_callback != NULL) {
+        (*interrupt_callback)();
+    }
+}
+
+void adc_init(void) {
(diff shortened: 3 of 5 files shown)


tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,51 @@
...
+#include <avr/interrupt.h>
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#define HW_TIMER_Hz     (30)
+#define SW_TIMER_SCALER (6)
+
+static volatile bool run_blink_led0 = true;
+static volatile bool run_blink_led1 = true;
```
amit-kumarh, line 12: Since this is not being changed in an ISR (hardware interrupt), it doesn't need to be designated as volatile

tutorials/super_blink/blink_config.h
amit-kumarh, line 1: lgtm

tutorials/super_blink/BUILD
amit-kumarh, line 1: lgtm

tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,51 @@
...
+    run_blink_led0 = true;
+}
+
+void sw_timer_callback(void) {
+    run_blink_led1 = true;
+}
+
+void sw_timer(void) {
+    counter += 1;
+    if (counter % (HW_TIMER_Hz / SW_TIMER_SCALER) == 0) {
```
amit-kumarh, line 25: wouldn't this simplify to `if (counter % 5) == 0`, running this every 5/30s of a second = 6 Hz?

tutorials/super_blink/super_blink.c
```diff
@@ -0,0 +1,51 @@
...
+
+static volatile bool run_blink_led0 = true;
+static volatile bool run_blink_led1 = true;
+uint8_t counter = 0;
+
+void timer0_callback(void) {
+    run_blink_led0 = true;
+}
+
+void sw_timer_callback(void) {
```
amit-kumarh, line 19: this does work, but in my personal opinion this pattern is a little over complicated, ex. there's not necessarily a need for a separate callback function like there is in a hardware CTC match, so it'd probably be a little bit easier to set `run_blink_led1` in the `sw_timer` function.

if you really wanted to trim it down, one way to do this would be to define a macro equal to HW_TIMER_Hz over your scaler (which I believe would be six) and increment your counter in the run_blink_led0 block, so you'd have something like

```
#define 5_HZ_SCALER 6
...
if  (run_blink_led0) {
    counter++
    ...
}

if  (counter % 5_HZ_SCALER == 0) {
   ...
}
```
which would also allow you to locally scope the counter to the main function instead of having it as global. not necessarily right or wrong, just how I would personally write it

amit-kumarh (changes requested):
great work! left some small comments.
questions:
1. not sure I entirely understand this question - the atmega only has two timers on it, so if you want to directly tie a timer to a gpio pin, you can only do that 2x. but if you have the interrupt and run some other code based on that, you run whatever code you want based on that.
2. usually don't need the static keyword in these instances. static does two things, if a global variable is static, that means it isn't visible in any other files (say if it includes our file). not really relevant here since this is all just one file. if it's in local in a function, it remains in memory (and thus holds its value) across function calls. this is usually used when we want to preserve some information or state between function calls (see throttle and BMS code for examples of this)

drewnotdrew:
1. I understand the two timers and how we can use interrupts - I was just confused why this is considered a pure hardware approach if the interrupts themselves are software and thus may create some small delay. My initial intuition was that a pure hardware timer would be directly tied to one pin and thus could not be configured. Definitely need to read more of the 16M1 documentation.
2.  This makes sense, but I think I'll have to use it more in practice for it to make complete sense. Ayush provided some additional context so I think I'm good for now.

drewnotdrew:
Added comments to all lines of the ADC library. I checked the commented version on main and understand everything from a high level, but I do have a few remaining questions:
3. Where are `ADCSRA`, `ADEN`, etc defined? Presumably these are numbers replaced by macros?

tutorials/super_blink/research/adc.c
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * REFS0 = REFSn: Selects reference for the ADC
+ */
+
+// Initializes the ADC on the MCU
+void adc_init(void) {
+    ADCSRA = (1 << ADEN) | (1 << ADPS2)
+             | (1 << ADPS0); // Sets the ADEN, ADPS2, and ADPS0 bits to 1.
+                             // Enables the ADC, sets the prescaler to 32
+    ADCSRB &= (1 << AREFEN); // What does this do? Clear all bits but AREFEN?
```
drewnotdrew, line 40: What is the purpose of this line? Clear all bits but `AREFEN`? Initial values for the bits in `ADCSRB` should be all 0, so clearing all bits but `AREFEN` (which is also 0) should set `ADCSRB` to `0x00`, no?

tutorials/super_blink/research/adc.c
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Check if an ADC conversion is done, and if so reads the result and sets it to
+// result
+int adc_poll_complete(uint16_t* result) { // Takes a pointer to a 16 bit int
+    // Checks if the ADIF bit is 0
+    if (!(ADCSRA & (1 << ADIF))) {
+        return -1;
+    } else { // If ADIF is 1
+        ADCSRA |= 1 << ADIF; // If ADIF is already 1, why do we need this?
```
drewnotdrew, line 60: On line 60 - this sets the `ADIF` bit to 1, correct? Based on the conditional this seems redundant.

tutorials/super_blink/research/api.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <stdint.h>
+
+// What is this?
+#define AS_VOLTAGE(raw) (float)raw*(0.004882813)
```
drewnotdrew, line 5: What is this macro used for?

tutorials/super_blink/research/adc.c
```diff
@@ -0,0 +1,87 @@
...
+// Reads the ADC register and set it to a 16-bit variable
+void adc_read_results(uint16_t* result) {
+    *result = ADC;
+}
+
+// Interupts the current ADC conversion
+void adc_interrupt_enable(
+    void (*callback)(void)) { // Takes a callback function as an input
+    ADCSRA |= (1 << ADIE); // Sets the ADIE bit to 1 to interupt the conversion
+    interrupt_callback = callback; // Interupt callback?
```
amit-kumarh, line 76: yep. if you look at the top of this file, the bit with ISR, that's the syntax Atmel gives us for defining the callback function. we just wrap it up nicely here so the user can pass it into the interrupt_enable function.

tutorials/super_blink/research/adc.c
amit-kumarh: Not quite sure on this one either to be honest - looking at the datasheet, it says that the ADIF is automatically cleared by the hardware, or can be alternately cleared by manually setting it to 1 (so somewhat counterintuitively, you write a 1 to set it to 0, also called write to clear). The reason why they do this is a little complex, but long story short it protects you from something called read-modify-write errors - essentially having the value of the register changing from under you while you're trying to clear a bit means you can accidentally clear stuff you didn't mean to. (don't worry if that didn't make total sense; I'm probably doing a bad job explaining it: [this](https://www.avrfreaks.net/s/topic/a5C3l000000UDgREAW/t066397) link has some more info)

Now, why we're manually clearing it I'm not too sure. It's possible the auto clearing wasn't working totally right (maybe because we don't have an interrupt attached?), as if the flag isn't cleared it looks like it won't let you run another conversion, so when we wrote the library the fix was to manually clear. Would be interesting to run a test without this line and see what happens!

amit-kumarh:
Questions:
1. Ah, I see. The process is pretty low level - it is literally telling the CPU to jump straight to our interrupt on the next instruction. but yeah there will theoretically be some tiny delay in there, but it would be much much less delay than whatever you get from doing all that modulo math though. You can tie a hardware timer straight to a GPIO pin for a blinky and have it set, clear, or toggle when it triggers, but obviously you're super limited in what you can do when you're not using any code (you couldn't like send a CAN message or something like that).
2. sounds good!
3. yep, they're macros defined in the AVR libraries we import

PR #249, opened 2023-10-06
Add executable permission to KiCAD Symbol Git Hooks + install scripts

amit-kumarh:
This should fix the issues folks have been having with the OEM symbol library not existing. 

Essentially, what went wrong was we have an install script that will add some scripts into `//.git/hooks` to run and generate the symbol library, but that script didn't give those new git hooks permission to be executed. This PR just adds a few lines to address that.

Also, just added a few lines to the monorepo README so that installing the git hooks is part of the documented setup procedure.


scripts/remove_kicad_git_hooks.sh
ayushchakra, line 1: why does this come up in the diff?

ayushchakra (approved):

scripts/remove_kicad_git_hooks.sh
amit-kumarh: marked it as executable

ayushchakra: ah didnt realize git captured that, but definitely makes sense that it does

PR #255, opened 2023-12-08
adding functionality to control cooling pump

Rixhee:


vehicle/mkvi/software/bms/bms.c
```diff
@@ -37,10 +37,45 @@ void timer0_isr() {
...
+    timer1_cfg.channel_b.output_compare_match = 0;
+    timer_init(&timer1_cfg);
+}
+
 void pcint0_callback() {
     bms_core.bspd_current_sense = !!gpio_get_pin(BSPD_CURRENT_THRESH);
 }
 
+static void control_cooling_loop(void) {
+    static int8_t tolerance = 5; // to be defined
```
ayushchakra, line 51: nit: this should be a macro since it doesn't change over time (ie `#define TOLERANCE (5U)`)

vehicle/mkvi/software/bms/bms.c
```diff
@@ -37,10 +37,45 @@ void timer0_isr() {
...
+    static uint16_t current_temp = 0;
+    static uint16_t previous_temp = 0;
+    static uint8_t duty_cycle = 0;
+    static int8_t min_duty = 13;
+    static int8_t max_duty = 85;
+    static int8_t scaling_factor = 1; // to be defined
```
ayushchakra, line 57: same as above ^

ayushchakra: in general, constants shouldn't be variables, they should be defined as macros

vehicle/mkvi/software/bms/bms.c
```diff
@@ -37,10 +37,45 @@ void timer0_isr() {
...
+
+    if (can_poll_receive_m162_temperature_set_3() == 0){
+        can_receive_m162_temperature_set_3();
+        current_temp = int(m162_temperature_set_3.d3_motor_temperature);
```
ayushchakra, line 61: any reason in particular on why these are ints?

ayushchakra: also you have to apply the scaling as described in the dbc to get the actual temperature value, not the decoded binary.

vehicle/mkvi/software/bms/bms.c
```diff
@@ -37,10 +37,45 @@ void timer0_isr() {
...
+    }
+
+    if (abs(previous_temp - current_temp) > tolerance && current_temp != 0) {
```
ayushchakra, line 64: you don't update `previous_temp` anywhere

vehicle/mkvi/software/bms/bms_config.c
```diff
@@ -49,10 +49,13 @@ timer_cfg_s timer1_cfg = {
     .timer1_mode = TIMER1_MODE_PHASE_CORRECT_PWM_10_BIT,
     .prescalar = CLKIO_DIV_8,
     .channel_a = {
-        .output_compare_match = 0x50,
+        .output_compare_match = 12,
```
ayushchakra, line 52: would be nice to include documentation on where the compare match values come from

vehicle/mkvi/software/bms/bms.c
```diff
@@ -225,6 +263,15 @@ int main(void) {
             //         loop_counter = 0;
             //     }
 
+            if (can_poll_receive_dashboard() == 0) {
+                can_receive_dashboard();
+
+                bool ready_to_drive = dashboard.ready_to_drive;
+                if (ready_to_drive) {
+                    control_cooling_loop();
```
ayushchakra, line 271: super nit: based on how you have it configured, we are only going to update the cooling loop PWM when we receive a message from the dashboard - would probably be better to decouple the timing of the systems (ie, in one logic block update the status of `ready_to_drive`, and in another block, implement the cooling loop logic based on the most recevent value received from the dashboard

vehicle/mkvi/software/bms/bms.c
```diff
@@ -225,6 +263,15 @@ int main(void) {
             //         loop_counter = 0;
             //     }
 
+            if (can_poll_receive_dashboard() == 0) {
+                can_receive_dashboard();
+
+                bool ready_to_drive = dashboard.ready_to_drive;
```
ayushchakra, line 269: why do you need to create a separate variable instead of directly using `dashboard.ready_to_drive`

vehicle/mkvi/software/bms/bms.c
```diff
@@ -37,10 +37,45 @@ void timer0_isr() {
...
+        duty_cycle = (int) (min_duty + current_temp * (max_duty - min_duty) / scaling_factor);
+        
+        if (duty_cycle > max_duty) {
+            duty_cycle = max_duty;
+        }
+        else if (duty_cycle < min_duty) {
+            duty_cycle = min_duty;
+        }
+
+        timer1_cfg.channel_b.output_compare_match = (duty_cycle/100) * timer1_cfg.channel_a.output_compare_match;
```
ayushchakra, line 74: doesn't the compare match value have to be an int, not a float?

ayushchakra:
overall, this looks amazing, awesome work! just made a few comments - one last thing, it would be really nice to include a debug CAN message that includes stuff like the current duty cycle, current and previous temps, and any other dynamic info for debugging

PR #256, opened 2024-02-09
V1 of J1772 and updated CAN controller

natsukisacks:


meaganlipsman:
Hi Suki! Awesome work on the board I really appreciate the effort you've put into this. I just have a couple of things that might be useful to change around a bit. First, just as a quality of life thing, could you add labels to the silk screen of the board to signify what each test pin is and where each microfit connector goes to? If you look on my board from last year, you can see an example of this in the "to switch" label. This will make it testing the board sooooo much easier. For the test points I actually didn't label them on the silk screen of my old board (my bad oops) but I think that would be super useful as well. Also, as a general note, since you have the space, it might be useful to spread out your components a bit more so that you can possibly avoid having so many vias and when you do have them they don't have to be so on top of each other. The main area I noticed this was around the program header. Besides that great work and thanks again!

drewnotdrew:
Great work so far! Here's a few things I'd recommend changing. Nitspicks are at the end of each section.

## Schematic
- Add no-connect labels here on pins 3 and 4. You can use the `q` shortcut to quickly add these. image
- Missing a jumper here on the can controller. I would add the same 2x1 header used on the can transceiver. 
image
- CAN_- and CAN_+ are the valid hierarchical labels here, make sure you use them inside the sub page for the CAN controller
image
- Example fix here for CAN_+:
image
- SCK here is not connected - either use a global label (less recommended) like MOSI, CS, or add a hierarchical label (recommended)
image

### Nits:
- Minor issue with a schematic boundary here:
image
- Bit of overlap on the right here:
image
- I know this is carried over from the previous rev, but I'd recommend changing the title of this block to something more descriptive, like it's functional purpose rather than the literal part.
image

## Layout
- For CAN termination, I'd use the 2x1 header footprint rather than my custom solder jumper. That footprint was designed for space-scarce boards and is less convenient than a header.
image
- Looks like a few 0805 footprints made it into your board. We've largely moved away from using 0805 in favor of 0603. Couple I see off the bat are the 30pf caps and 10k resistors. Leds can be either 0603 or 0805
image

### Nits:
- Something that might help with the aesthetics of your board is if you aligned all of your connectors. Except for the programming header, generally connectors should be placed on the edges of the board. You could move them all to one edge. If you're looking for a good example, I'd take a look the the rkh.
image

Haven't fully finished looking at the layout - I'll add more comments for anything I see later today.

drewnotdrew (changes requested):
Second pass on the layout looks a lot cleaner! Nice work. 

Schematic:
- The ERC is complaining about component annotations. Make sure you re-annotate after adding an new components and update the layout accordingly.
- For the CAN controller you are using the hierarchical labels `CHARGER_CAN_+` and `CHARGER_CAN_-`, but inside `CAN_+` and `CAN_-` inside the module. Make sure these are the same - the erc will complain if not. 
- Same with the SPI lines to the CAN controller. The label used at the top level schematic must match the label used in the sub-schematic.

Layout:
- Looks like a few 0805 parts snuck through - check the 100pf C4 capacitor and 30pf C3 capacitor, as well as double-check every other passive.
- Move the crystal as close to the MCU as possible. The signal lines from the crystal should be prioritized over any other traces.
- Same with the CAN controller crystal. It could be nudged a little closer.
- Nit: Decoupling caps should also be placed as close to their respective components as possible. 
For example:
Prefer: 
image
Over:
image
- If you have time: Add power leds for 5V and 12V.
- The car CAN lines are disconnected - resolving the issues with the hierarchical labels in the schematic will update the rats nest to display where you are missing connections. When you route these, keep the traces as close to each other as possible.
- Same with the charger CAN lines. 
Prefer:
image
Over:
image
- The `CHARGER_CAN_-` line is 15mil, make sure it's 8mil.

meaganlipsman:
Hi Suki, great job and major +1 to what Drew said about this version's layout being a lot cleaner! If you have time, I think it could be really great to add the silkscreen labels to the board, this will make it exponentially easier for us to test. This is how they were done on my board last year: 
image

You can do this by clicking on the silkscreen layer and then the text button. This is an example of how this could look on your current version of the 
image
I through this together quickly, but if you are able to add in the labels it would be ideally a tad cleaner but it doesn't have to be anything fancy. Also I would breifly think about where your test points are placed, If you look on the 3D viewer you can see that they might be placed a little close to the header pins and therefore might make it more difficult than it has to be to test. 
image
You can access the 3D viewer by going to view >> 3-D viewer. They don't need to be moved drastically but having them slightly more spread out could be good for neat labeling purposes as well! 

Also in your layout there seems to be a few 5V traces that are not connected? not sure if this is on purpose but just wanted to point it out in case

Overall this looks great, awesome work!

drewnotdrew:
Your recent changes look great! Here's a few last recommended changes:
- Somewhere along the way pins 3 and 4 on the 6 pin microfit got removed. Add them back like how they are in rev 1:
image
- Add labels to the termination resistor headers. These could be as simple as "Car Term" and "Charger term".
- The can transceiver (MCP2561) should be the soic-8 package rather than the DFN8. This will make soldering much easier.
- We've recently unofficially [standardized to 0603 LEDs](https://coda.io/d/MKVII-Documentation_dLTyp9cQyXS/Standardization_suN14#Led-Color-Code_tuzZl/r7). 
- Add your and Meagan's names!

drewnotdrew (approved):

PR #258, opened 2024-02-28
Add MKVI BMS Voltage Task

drewnotdrew:
Adds voltage task from `amit/bms-code-dev`.

All voltage task fault states should be tested, and the code requires a little bit of cleanup before merging.

Current state: Builds and functional, not fully tested.


Rixhee:
Fault check list for testing:
- [ ] UNDERVOLTAGE
- [ ] OVERVOLTAGE
- [ ] CSC_MIA

PR #259, opened 2024-02-28
Add MKVI BMS Temperature Task

drewnotdrew:
Adds temperature task from `amit/bms-code-dev`.

All temperature task fault states should be tested, and the code requires a little bit of cleanup before merging.

Current state: Builds and is functional, not fully tested.


drewnotdrew:
- [ ] Undertemperature
- [ ] Overtemperature

PR #262, opened 2024-03-07
Add PCIE x1 Passthrough Breakout Board

drewnotdrew:
This is a PCIE breakout board with a board edge connector and socket connector with all pins broken out. There is also a panelized version with variations of the riskiest part of the board - the PCIE board edge connector.

image


drewnotdrew:
Panelization complete. PCIE connector dimensions appear quite standard across the board, so the only variation between each board is some adjustment to the ends of the pads and the surrounding silkscreen.
Ready to merge.
image
image

Eemac (approved):

PR #263, opened 2024-03-08
Add MKVI Cooling Board

ishanporwal:
Cooling board controls radiator fan and water pump.


drewnotdrew (changes requested):
Great work on this board! Here's a recommended few changes, as well as a few things you may want to clarify in documentation. Nitpicks and are largely up to personal preference and you can choose to ignore them.

### General
- Split the hardware and firmware for your board into two PRs - it'll be easier to review each independently.
- On the firmware branch, make sure to format your code with clang. Unformatted code is causing one of the CI checks to fail.
- Remove the `.lck` files from this PR.

### Schematic
Everything looks good here, except for the power circuitry of the fan. Do you have the resources you used to make it? I believe the pins may be swapped on the transistor symbols. A simpler alternative may be to instead use a single low-side drive mosfet. We have plenty of `SSM3K333R` and you can reference the nostril dashboard for how it's used, if you choose to switch.
- Nit: Add a 2x1 header in series with the 120 ohm termination resistor. Not necessary, but it makes it a whole lot faster to add and remove the termination resistor instead of soldering/desoldering.
- Nit: What is the purpose of this global label? Remove it if does not have an explicit purpose.
image
- Nit: `Water Signal` is a bit ambiguous. I would use `water_pump_pwm` or `pump_pwm` here.
- Nit: A 160 ohm resistor will make D2 very bright. I would switch to this to a 330 ohm resistor.

### Layout
- There's a few 0805 components in the layout - make sure these are 0603. A few I see are the LEDs, 30pf capacitors, and some of the resistors. I would make one pass through all the passives to make sure you don't have any more 0805 components.
- What is this label referring to? You may want to clarify this in your documentation.
image
- Nit: The `5KW LED` label appears to reference the fan power LED, and the `Fan PWR LED` appears to reference a debug LED. I would change these to `Fan PWR LED` and `Debug LED` respectively.
- Nit: This label likely won't show up well. JLC specifies text must be 1mm tall, but you can push it to 0.7-0.8mm.
image

drewnotdrew (approved):
lgtm. Will merge once it's ordered.

PR #265, opened 2024-03-18
Update BMS Fault Handling

drewnotdrew:
Previously, faults no two faults could be present at the same time on the BMS core firmware. Now, multiple faults can exist at the same time, and all are transmitted over CAN. Canviewer has been updated to decode the `bms_fault_code` and list all present faults.

Currently untested as I do not have access to the hardware right now.


vehicle/mkvi/software/bms/utils/fault.h
```diff
@@ -1,7 +1,47 @@
...
+ */
+int check_fault_state(void);
+
+/**
+ * Clear all BMS faults.
+ *
+ * Use with caution. However, any faults caused non-transient effects will
+ * trigger faults again.
+ */
+void clear_all_faults(void);
```
drewnotdrew, line 47: Added this function with the idea of tying it to a CAN message that we can send from a laptop plugged into the car, but that seems a bit unsafe. Even having a function that clears all of the faults seems unsafe. Thoughts on keeping this, or better ways of clearing faults?

My thinking with clearing faults is that if we clear all faults and the segments are in an unsafe state, we'll see faults on the next loop and will never open the BMS relay. Then again, maybe we should do a bit of filtering to ensure transient spikes in data don't cause fault states.

drewnotdrew:
image
Tested and working when manually setting all faults. `bms_fault_code` should eventually be moved to the fault section of the vehicle states column.

PR #266, opened 2024-03-21
Add Revision D of IMU

drewnotdrew:
Changes:
- Switched programming and IMU SPI lines to the `spi_a` lines
- Updated LED resistor value 91->330 ohms
- `OE` is now connected to `VCCA` instead of `VCCB`,  from the TXB0104 datasheet
- Lengthened pads of the ICM20948 and TXB0104 for easier rework
- Added test points for `MOSI`, `MISO`, and `SCK`.
- Reduced all signal trace widths to 3.5mil, per JLC specifications for >4 layer boards
- Connected `INT1` to MCU from the IMU
- Added thermal pads back to the MCU and CAN transceiver
- Panelized 4x4 for cost


mkazc:
Looks good to me, and the changes you've made make sense. My only worries are related to spacing, but given the size of the board, it was done well. I would personally return the pads of the ICM20948 and TXB0104 back to their original sizes (if they were manufacturer footprints) given those constraints in relation to the solder mask. If the manufacturer says that it's okay, definitely feel free to continue with that.

drewnotdrew:
The widths of the pads were not changed, only the lengths so there shouldn't be any issues with the solder mask. If you think there would still be issues with this let me know.

Here's the new and old footprints side by side:
Old:
image
New:
image

The difference is most apparent in the 3D viewer:
Old:
image
New:
image

drewnotdrew:
Now panelized!
image

mkazc (approved):
Looks good! Great work!

PR #267, opened 2024-03-21
Create Carrier Board Fix

mkazc:
Created Carrier Board to fix current issues on segment, in order to avoid working with a live segment, along side with other changes. Things to note:

- SDA / SCL pins are swapped - only use the the connectors we bought, not the fixed ones
- Throughholes on the left side are meant to help with mounting
- The MicroFit 9x2 connector is part of the fix, goes to cell pads on Power Board


drewnotdrew (commented):
Looks good to me. My only suggestion is it may be worth rotating the MicroFit so it's easier to connect/disconnect when the carrier board is mounted.
carrier_1

It's tight, but this might work:
image

Or even:
image

drewnotdrew:
Ready for final review. Changes:
- Rotated connector, rerouted traces
- Changed pinout of microfit for trace routing
- Added labels to most pins on the microfit
- Reduced overall width to 29.6mm, the same width as a CSC.
image

mkazc:
The angled connector is perfect! Great work

drewnotdrew (approved):

PR #268, opened 2024-03-22
Add MKVII BMS DA Board

agathaford:
First rough draft of mkvii da module layout and updated schematic.


mkazc:
This looks amazing! The most of the things to look out for are found by doing a Design Rules Check (DRC). This includes traces too close to any edges, and the Edge Cut being slightly broken (I believe on the left side there's an extra line). I would change the MUX output in-line resistor to maybe 100 Ohms unless told otherwise ^^

PR #271, opened 2024-04-06
BMS I2C Error Detection for DA Board Mixes

ayushchakra:
This PR is still a WIP, just putting it up for visibility.


vehicle/mkvi/software/bms/bms.yml
```diff
@@ -301,49 +305,66 @@ publish:
           scale: 0.0001
           offset: 0
 
-  - name: cfg_reg_a
```
ayushchakra, line 304: this diff is going to look weird, basically, we had an old `cfg_reg_a` bms message that's an artifact of the very beginning of bms testing that I deleted and changed to `bms_mux`

PR #279, opened 2024-09-11
Add IMU Code

drewnotdrew:
**Current features:**
- Health heartbeat LED @ 1 Hz
- XYZ accel over CAN @ 217Hz
- XYZ gyro over CAN @ 217Hz
- Automatic CAN arb id setting based on solderable pull ups

**SPI Library:**
- The SPI library now supports using the main SPI bus (MISO, MOSI, SCK) or the alt SPI bus (MISO_A, MOSI_A, SCK_A). As always, programming is always performed with the alternative bus. 

**Future:**
- FIFO use for faster and more efficient data read
- Magnetometer data
- Sensor fusion and on-sensor or on-mcu quaternion calculation
- Additional on-sensor filtering if needed


vehicle/mkvi/software/imu/imu.yml
```diff
@@ -0,0 +1,77 @@
+name: imu
+publish:
+  - name: imu_accel
```
ayushchakra, line 3: just doing some quick math, 48 bits * 326 messages / sec * 2 messages roughly equals 31 kbps. probably isn't a concern based on our current can bus (I think you said we we're at 25% utilization at some point?), but is something we should probably keep in mind if we have an array of IMUs / add other high frequency messages

also have you tested that your can messages are actually being sent at that frequency? id be impressed if the atmega can actually handle that since the can functions (and I think SPI?) functions are blocking

.github/workflows/kicad.yml
```diff
@@ -47,10 +47,10 @@ jobs:
       #     out_dir: kicad-artifacts
       #     versioning: false
 
-      - name: comment
-        uses: marocchino/sticky-pull-request-comment@v2
-        with:
-          path: build/comment.md
+      # - name: comment
+      #   uses: marocchino/sticky-pull-request-comment@v2
+      #   with:
```
ayushchakra, line 52: is this still broken

vehicle/mkvi/software/imu/icm20948/icm20948.c
```diff
@@ -0,0 +1,71 @@
+#include <stdint.h>
```
ayushchakra, line 1: nit: it seems that Amit started a convention of throwing common drives (ex ltc6811) into the vehicle/common dir, probably worth throwing this in there (unless you anticipate it could change, causing previous firmware versions to break)

vehicle/mkvi/software/imu/icm20948/icm20948.c
```diff
@@ -0,0 +1,71 @@
...
+
+    switch_register_bank(BANK_3);
+    // Read data
+    icm_read_register(I2C_SLV4_DI, &_rx_data);
+    // icm_read_register(EXT_SLV_SENS_DATA_00, &_rx_data);
+    // _can_print(_rx_data);
+}
+
+void write_magnetometer(uint8_t register_, uint8_t tx_data) {
+    ;
```
ayushchakra, line 70: should there be a todo here?

vehicle/mkvi/software/imu/icm20948/icm20948.c
```diff
@@ -0,0 +1,71 @@
...
+
+    switch_register_bank(BANK_0);
+}
+
+void read_magnetometer(uint8_t register_, uint8_t* rx_data) {
+    switch_register_bank(BANK_3);
+    uint8_t _rx_data;
+
+    // Write the magnetometer addr for reading
+    icm_write_register(I2C_SLV4_ADDR, (I2C_READ << 7) | MAGNETOMETER_ADDR);
```
ayushchakra, line 47: this might be my unfamiliarity with the chip, but couldn't this just be done once in `init_magnetometer`

vehicle/mkvi/software/imu/icm20948/icm20948.h
```diff
@@ -0,0 +1,256 @@
...
+ * @param _register (uint8_t) The register to write.
+ * @param tx_data (uint8_t) The data to write.
+ */
+void icm_write_register(uint8_t _register, uint8_t tx_data);
+
+/**
+ * ICM register banks.
+ */
+#define BANK_0 0
+#define BANK_1 1
```
ayushchakra, line 227: nit: make this an enum?

vehicle/mkvi/software/imu/imu.c
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Timer 1 for the LED heartbeat
+ */
+volatile bool led_heartbeat = true;
+void timer_1_isr(void) {
+    led_heartbeat = true;
+}
+
+uint8_t test = 0;
```
ayushchakra, line 42: nit: delete

vehicle/mkvi/software/imu/imu.c
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * Update the arb id of an IMU based on soldered pull up jumpers.
+ *
+ * Allows for multiple IMUs running the same firmware on the same CAN bus.
+ */
+void update_can_arb_id(void) {
```
ayushchakra, line 48: not in the scope of this PR, but in the future, we should expand the functionality of our CAN API to handle boards with multiple can IDs for the same message - that way, when our DBC gets assembled, we can automatically detect if there's id collisions

vehicle/mkvi/software/imu/imu.c
```diff
@@ -0,0 +1,168 @@
...
+    // Pull up inputs
+    gpio_set_mode(bin_id_0, INPUT);
+    gpio_set_mode(bin_id_1, INPUT);
+    gpio_set_mode(bin_id_2, INPUT);
+
+    // Update ids
+    imu_accel_msg.id
+        += (0x10
+            * (!gpio_get_pin(bin_id_0) + (!gpio_get_pin(bin_id_1) << 1)
```
ayushchakra, line 57: nit: just compute the multiplying factor once and apply it to each

vehicle/mkvi/software/imu/imu.c
```diff
@@ -0,0 +1,168 @@
...
+    // Initialize magnetometer
+    // init_magnetometer();
+}
+
+/**
+ * Read accel data and place in CAN structs
+ */
+void read_accel_data(void) {
+    uint8_t accel_data[2] = { 0x0, 0x0 };
+    icm_read_register(ACCEL_XOUT_L, &accel_data[0]);
```
ayushchakra, line 115: from my very very quick scan, I think the chip supports multibyte reads?
Screenshot 2024-10-05 at 10 52 48 AM

.github/workflows/kicad.yml
drewnotdrew: This is the comment from Github Actions that's currently broken. It can be added back if/when we fix that action and find a place to store files (e.g. Digital Ocean).

vehicle/mkvi/software/imu/imu.yml
drewnotdrew: I agree, and I think given our current hardware it'd be wise to keep it all on a separate bus, which is what the harness I sent acts as.

I didn't measure anything rigorously, I only adjusted the main timer compare match such that the reported CAN bus load from `canbusload` was around 9% per IMU. I sent the tag connect with the IMUs so this can be updated if need be. The same applies if we get garbage data when reporting data too fast.

In terms of sending data in the 200-400Hz range, I believe the Atmega and it's CAN peripheral can handle it. Reporting at 300Hz gives the Atmega 3.3ms to perform 12 8-bit SPI reads @ 4MHz and send 3 CAN messages @500KHz. So I'd estimate around  (3.3ms - (12 * 8)/4M - (64 * 3)/500k) = 2.9ms of overhead. That of course with a lot of assumptions - no overhead for CAN messages or SPI, etc.

vehicle/mkvi/software/imu/icm20948/icm20948.c
drewnotdrew: The general API should stay the same. I've thrown it in `vehicle/common`.

vehicle/mkvi/software/imu/icm20948/icm20948.h
drewnotdrew: Added a todo. Don't have the hardware so I don't want to change too much without testing.

vehicle/mkvi/software/imu/imu.c
drewnotdrew: Yes, and you can also use the built in FIFO buffer in the IMU to optimize even further. The SPI interface wasn't a bottleneck so I didn't get to it. Could be more useful in the future with a more powerful MCU.

vehicle/mkvi/software/imu/icm20948/icm20948.c
drewnotdrew: There's some quirks with the magnetometer that I wasn't able to fully resolve. I could read the device ID but no more. I was copying Ian's register writes so I'm not quite sure why this was in every magnetometer read.

ayushchakra (approved):

PR #282, opened 2024-09-17
Add Can Print

drewnotdrew:
See `projects/can_print/README.md` for full documentation.

## can print

`can print` is a tool for print-based debugging over CAN. All static inputs of a `can_print` message are processed at _compile time_ using C function-like macros. Accordingly, an initial configuration of `can_print` with a single print message adds approximately 400B to an image, and each additional instance of a message adds 20B. `can_print` messages are set as the lowest priority of standard 11-bit arbitration IDs via the `0x7FF` ID.


Eemac (approved):
w can print

PR #294, opened 2024-09-30
Add Can Print

Eemac:


bazel/defs.bzl
```diff
@@ -291,6 +291,18 @@ def cc_firmware(name, **kwargs):
         data.append(btldr + ".hex")
         defines.append("BTLDR_ID=" + ECUS[btldr_ecu_name]["btldr_id"])
 
+    is_can_print = False
```
ayushchakra, line 294: just curious, what's the motivation behind also having cc_firmware be able to launch the print decoder? couldn't we just have a wrapper function that calls cc_firmware and the py_binary that you created to have the same effect?

examples/can_print/BUILD
```diff
@@ -0,0 +1,61 @@
+load("//bazel:defs.bzl", "cc_firmware")
+load("//bazel/tools:defs.bzl", "can_api_files", "dbc_gen")
+
+exports_files([
+    "can_print_example.yml",
+])
+
+# Used for debugging with compiler preprocessor output.
+cc_binary(
```
ayushchakra, line 9: still needed?

projects/can_print/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+First, add the `can print` header to a source file and use the `can print`
+macro.
+
+```
+// example_file.c
+#include "projects/can_print/can_print.h"
+
+// Example usage
+can_print("regen_tq", -32085, 3);
```
ayushchakra, line 27: nit: put the include in ur example

bazel/defs.bzl
drewnotdrew: I was in the process of learning Bazel rules and started with an implementation that was less than optimal. I also wanted there to be no additional boilerplate in BUILD files. 

I think you're right - I should be able to add a wrapper that calls both. I'll give it another shot, but if it doesn't work the intended use of `can print` still works - that is run the decoder in a separate shell and flash firmware while developing.

examples/can_print/BUILD
drewnotdrew: Yes. Because of the heavy use of C macros viewing the `-e` output is invaluable. I'll keep it since it's a rule that's out of the way.

drewnotdrew:
Remaining todo:
- [ ] Add arb id setting to documentation
- [ ] Debug CI
- [ ] Rearch main can print Bazel rule

bazel/defs.bzl
drewnotdrew: I'm remembering why bundling them together in a macro was difficult - passing some arguments to one rule and the rest to another rule isn't straightforward, at least in simple macros. I believe it's possible with more advanced Bazel features, but I think the simplest solution would instead be a `can print` config file.

For now I'll file in the errata until we run into issues or I learn more Bazel.

ayushchakra: sg

ayushchakra (approved):

PR #295, opened 2024-10-01
Fix CI

drewnotdrew:
- Formatted all Bazel files
- Removed some HiTL deps
- Updated paths headers for examples that were moved
- Removed and added back Bazel cache CI to get rid of failing cache
- Commented out HiTL pytest tests. These will need to be debugged if HiTL is ever picked up again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: ci
 
-on: 
+on:
   pull_request:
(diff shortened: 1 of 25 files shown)


drewnotdrew:
Interesting. Force-pushing back to main automatically closes the PR.

ayushchakra (approved):

PR #299, opened 2024-10-09
Summer2024/e leads

Rixhee:
For drew to review my board🗡️


drewnotdrew (commented):
## Schematic
Avoid connecting LEDs directly to signal lines. Adding status LEDs here loads the signal lines and leads to higher rise and fall times since the impedance of the signal line is lowered. If you would like to keep these LEDs, they are best buffered by a MOSFET or opamp.
image

Similarly, avoid placing LEDs directly on the shutdown circuit. While each LED in this case will draw around 3mA, if every board on the car had LEDs directly on the shutdown circuit, certain parts of the shutdown circuit would have to be rated for tens of milliamps. Buffering with a MOSFET or opamp would also work here.
image

Cell voltages here should be buffered by an opamp to prevent the resistor divider from affecting the cell voltage. Additionally, filtering on the cell voltage wouldn't hurt - check the BMS CSC schematic for an example.
image
to
image
A monolithic package with multiple opamps may be more suitable here with this many opamps.

### Nitpicks:
Since the MCU handles all of the logic of the shutdown relays, these relays could be combined into a single relay, no? The function would be the same, except the logic ANDing the `analog` and `shutdown` nodes would be handled in firmware rather than hardware. 
image

Looks like this is a decoupling capacitor. Either place it near the component in the schematic or make a note.
image

When possible, keep symbols from intersecting for readability. Also keep the test point reference designator so it can be referenced precisely without ambiguity. 
image

Personal preference here: place a solder jumper or header with a shunt in series with the termination resistor and populate the resistor, for posterity.
image

If you have additional GPIO pins, always connect the auxiliary pins of an IC for future extensibility. Examples of these types of pins are enable, reset, standby, etc.
image

For readability, use the 5V power symbol over a regular net label.
image image

Since these are identical, they could be put in hierarchical sheet to link changes.
image

Overall looks great! I'll review the analog section and board layout tomorrow.

drewnotdrew:
## Schematic Cont.

### Analog Section
Is there a reason why the analog inputs are amplified and then divided? There may be something I'm missing here, but it seems like the amplifiers and dividers could be omitted here. Perhaps you meant to put the dividers on the inputs to the comparator?
image

The common mode input range (CMIR) of the LM397 is V_s - 1.5V, which would likely be exceeded with the preceding circuit. I'm not sure why comparators have this limitation, but it's something to watch out for. I've also seen this limit on an AD comparator.
image
For reference, the CMIR of the MCP6001:
image
The circuit shouldn't catastrophically fail, since the LM397 inputs pins are rated up to 30V, but the comparator output will fail to switch when expected.

**Nitpicks**
For any analog section I find it most help to label the expected conditions at that particular stage. This goes for any stage - RC filter, comparator, amplifier, etc. This not only double checks your work for the expected behavior of the circuit, but also helps you identify areas where a signal may rail. For example:
image
In more demanding cases, you could extend this with current, capacitance, etc.

**Comments**
To extend the comparator circuit a bit further, look into comparator hysteresis. If your input signals are particularly noisy, hysteresis prevents an indecisive comparator output.
image

drewnotdrew:
## Layout
Nice work on the layout - besides decoupling capacitor placement on the TC72 I don't see any major issues that would keep this board from working.

The TC72 should have a decoupling capacitor placed closed to it's power input. I believe this is C108?
image

The placement of the decoupling capacitors for the MCU and CAN transceiver, as well as the crystal could be placed closer to their respective components. For instance:
image
to
image
Manufacturers often specify placement of decoupling capacitors as close as the courtyards allow. For example, here's a snippet from a TI datasheet:
image

### Nitpicks
Always add mounting holes, even if it's not apparent when designing the board that you'll need them. M3 work well for mid-sized PCBs, and there's plenty of options in the built-in Kicad library.

Traces can be tidied to no end, so take this nitpick especially lightly. I prefer to space my traces evenly across the whole trace and transition to a consistent spacing if need be. This goes for the spacing of everything - traces, vias, components.
Ex:
image
to
image
Ultimately it doesn't matter, the function is the same and it takes extra time to space out traces.

PR #312, opened 2025-01-30
Add Mk. VII BSPD Firmware/KiCAD Files

H-TejadaDeras:
C File

- Added heartbeat led to update_leds function
- Added Heartbeat LED to gpio_set_mode declarations
- Moved pcint0_callback() into BSPD "while" loop
- Added RC_Circuit_Status and BRAKE_PRESS_SENSE_FILT to send can loop

H File
* Added RC_TIMER_STATUS, BRAKE_PRESSURE_SENSE_FILTERED, HEARTBEAT_LED pin declarations
* Added comments to explain new pin declarations
* Modified a few comments to fit my personal preference
* unvommented BSPD_SHUTDOWN_SENSE pin dec
* Updated MOTOR_5kW_LED pin dec to new pin (PB5 → PD5)

Yaml File

- Updated file to have new CAN outputs (heartbeat and brake_press_sense_filt)


vehicle/mkvii/software/brakes/BUILD
amit-kumarh, line 1: :+1:

vehicle/mkvii/software/brakes/bspd.yml
```diff
@@ -0,0 +1,64 @@
...
+    freq_hz: 100
+    signals:
+
+      # Shutdown Sense/BSPD Status
+      - name: ss_bspd
+        slice: 16 + 0
+        unit:
+          type: bool
+          values:
+            t: CLOSED/NRM. CND.
```
amit-kumarh: I'd have to double check, but my guess this might mess with the the CANViewer Shutdown Sense highlighting. Not a deal-breaker, but something we should check on.

vehicle/mkvii/software/brakes/bspd.yml
```diff
@@ -0,0 +1,64 @@
...
+            f: OPEN/FAULT
+
+      # Heartbeat Status
+      - name: heartbeat
+        slice: 16 + 1
+        unit:
+          type: bool
+          values:
+            t: Working Status 1 (has to be flipping between status 1 and 2 to be working)
```
amit-kumarh: nit: This message might be a little bit long to be flying by us when live debugging

vehicle/mkvii/software/brakes/bspd.yml
```diff
@@ -0,0 +1,64 @@
...
+            f: Working Status 2 (has to be flipping between status 1 and 2 to be working)
+
+      # Brake Light Status
+      - name: brake_gate
+        slice: 17 + 1
```
amit-kumarh: Please make sure your slice sizes are consistent with your data type

vehicle/mkvii/software/brakes/bspd.yml
```diff
@@ -0,0 +1,64 @@
+name: bspd
+publish:
+  - name: bspd
+    id: 0xb
+    freq_hz: 100
+    signals:
+
+      # Shutdown Sense/BSPD Status
+      - name: ss_bspd
+        slice: 16 + 0
```
amit-kumarh: I'd recommend moving your integer values to the top to avoid unaligned writes.

vehicle/mkvii/software/brakes/bspd.c
```diff
@@ -0,0 +1,124 @@
...
+    gpio_enable_interrupt(MOTOR_CURRENT_SENSE);
+    gpio_enable_interrupt(BSPD_LL);
+
+    ////////////////////////////// BSPD LOOP /////////////////////////////
+    for (;;) {
+        // BLTDR loop poll function
+        updater_loop();
+
+        // Get updated digital inputs
+        pcint0_callback();
```
amit-kumarh: Please move back to configuration section above the superloop

vehicle/mkvii/software/brakes/bspd.c
```diff
@@ -0,0 +1,124 @@
...
+    // Update CAN struct with new board logic values
+    bspd.brake_gate = !!gpio_get_pin(BRAKELIGHT_LL);
+    bspd.bspd_5kw = !!gpio_get_pin(MOTOR_CURRENT_SENSE);
+    bspd.ss_bspd = !gpio_get_pin(BSPD_LL);
+}
+
+// Check whether an LED needs updating, and if so, change its state
+void update_LEDs(void) {
+    // Update Brake Light LED on the PCB
+    if (bspd.brake_gate == true) {
```
amit-kumarh: nit: Don't need `== true` here, since `bspd.brake_gate` is already a `bool`. Same with line 61.

vehicle/mkvii/software/brakes/bspd.c
```diff
@@ -0,0 +1,124 @@
...
+
+    // Update BSPD Status LED/BSPD Trip LED on the PCB
+    if (bspd.ss_bspd) {
+        gpio_set_pin(BSPD_TRIP_LED);
+    } else {
+        gpio_clear_pin(BSPD_TRIP_LED);
+    }
+
+    // Update Heartbeat LED on the PCB; Actions chosen to make LED blink every cycle (showing heartbeat)
+    if (bspd.heartbeat == true) {
```
amit-kumarh: Please fix the heartbeat LED to blink at a human-readable interval using your 100hz CAN timer

amit-kumarh (changes requested):
Great Job!

H-TejadaDeras (commented):
Version 3 (Preliminary Version 2) - 02-05-2025

c file

* Moved get analog inputs out of send can loop (to clean up code)
* Added more comments to code (// get initial digital signals for pcint0_callback function)
* Moved and fixed Heartbeat Counter LED logic to Update LEDs Function
* Removed == True from if statement arguments for code cleanliness

yaml File

* Corrected data slice code
* Moved Brake Pressure Data (unfilt. and filt) to top for optimization reasons (runs faster when bytes don’t have to be split over different data segments)

H-TejadaDeras:
Version 3 (Preliminary Version 2) - 02-05-2025

c file
- Moved get analog inputs out of send can loop (to clean up code)
- Added more comments to code (// get initial digital signals for pcint0_callback function)
- Moved and fixed Heartbeat Counter LED logic to Update LEDs Function
- Removed == True from if statement arguments for code cleanliness
- Restored original location of pcint0_callback function (outside superloop)

yaml File

- Corrected data slice code
- Moved Brake Pressure Data (unfilt. and filt) to top for optimization reasons (runs faster when bytes don’t have to be split over different data segments)

vehicle/mkvii/software/bspd/.vscode/launch.json
amit-kumarh, line 1: Please remove the `.vscode` folder from your pull request

vehicle/mkvii/software/bspd/BUILD
```diff
@@ -27,6 +27,6 @@ cc_firmware(
 
 can_api_files(
     name = "can_api",
-    dbc = "//vehicle/mkvi:mkvi.dbc",
+    dbc = "//vehicle/mkvii:mkvii.dbc",
```
amit-kumarh, line 30: Good change here! Make sure you also change `//vehicle/mkvii/software/BUILD` file to point at your new YAML file too

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -31,13 +31,26 @@ void pcint0_callback(void) {
     // Update CAN struct with new board logic values
     bspd.brake_gate = !!gpio_get_pin(BRAKELIGHT_LL);
     bspd.bspd_5kw = !!gpio_get_pin(MOTOR_CURRENT_SENSE);
-    bspd.ss_bspd = !gpio_get_pin(BSPD_LL);
+    bspd.ss_bspd = !gpio_get_pin(BSPD_SHUTDOWN_SENSE);
+}
+
+// // Used to update the ss_bspd can signal to OPEN (but not back to CLOSED) when BSPD faults
+void update_ss_bspd_can(void) {
+    if (skip == false) {
```
amit-kumarh: I like the thinking here to ensure the BSPD system is properly being latched, but in this case because this is a sensing circuit whose purpose is to tell us to actual state of the shutdown circuit, it'd probably be best to just report the actual state from our sense circuit without adding intermediate code in the picture.

In a scenario where our hardware BSPD fails to latch, we wouldn't want the code to fail to report that the BSPD went back to closed again!

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -50,21 +63,25 @@ void update_LEDs(void) {
...
-    // Update Heartbeat LED on the PCB; Actions chosen to make LED blink every cycle (showing heartbeat)
-    if (bspd.heartbeat == true) {
-        gpio_clear_pin(HEARTBEAT_LED);
-    } else {
-        gpio_set_pin(HEARTBEAT_LED);
-    }
 }
 
+// Triggers Heartbeat LED every 0.5 s (based on 100 Hz CAN Signal)
+void update_heartbeat_LED(void) {
```
amit-kumarh, line 68: Nicely done!

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -92,33 +109,44 @@ int main(void) {
...
 
-        // Get updated digital inputs
-        pcint0_callback();
+        // Get updated analog inputs
+        bspd.brake_pressure = adc_read(BRAKE_PRESSURE_SENSE);
+        bspd.brake_pressure_filtered = adc_read(BRAKE_PRESSURE_SENSE_FILTERED);
+        bspd.opamp_timer_rc_circuit_status = adc_read(RC_TIMER_STATUS);
+
+        // // Correct ss_bspd can signal
+        // update_ss_bspd_can();
```
amit-kumarh: If this function is dead, please remove this comment

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -92,33 +109,44 @@ int main(void) {
...
 
     // Attach Pins to interrupt handler (assuming on rising/falling edge)
     // These are all PB registers which are mapped to the pcint0_callback
     // function
     gpio_enable_interrupt(BRAKELIGHT_LL);
     gpio_enable_interrupt(MOTOR_CURRENT_SENSE);
-    gpio_enable_interrupt(BSPD_LL);
+    gpio_enable_interrupt(BSPD_SHUTDOWN_SENSE);
+
+    // Gets initial analog inputs
```
amit-kumarh, line 116: nit: these are digital inputs, not analog. Your analog inputs you are getting with `adc_read` in your superloop

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -50,21 +63,25 @@ void update_LEDs(void) {
...
 
-    // Update Heartbeat LED on the PCB; Actions chosen to make LED blink every cycle (showing heartbeat)
-    if (bspd.heartbeat == true) {
-        gpio_clear_pin(HEARTBEAT_LED);
-    } else {
-        gpio_set_pin(HEARTBEAT_LED);
-    }
 }
 
+// Triggers Heartbeat LED every 0.5 s (based on 100 Hz CAN Signal)
```
amit-kumarh: This comment is not accurate since this is not being called in your `send_can` block

vehicle/mkvii/software/bspd/bspd.c
```diff
@@ -50,21 +63,25 @@ void update_LEDs(void) {
         gpio_clear_pin(MOTOR_5KW_LED);
     }
 
-    // Update BSPD Status LED/BSPD Trip LED on the PCB
-    if (bspd.ss_bspd) {
+    // Update BSPD Status LED/BSPD Trip LED on the PCB (can only turn on)
+    // Inverted to ensure appropriate behavior. If BSPD_LL is low (Relay is Opened) -> Trip LED Turn On
+    if (!gpio_get_pin(BSPD_LL)) {
```
amit-kumarh, line 58: Similar note here; I don't quite remember if we have this light and a separate hardware only BSPD SS light?

I'd say we definitely want to have as a first priority shutdown sensing where we are _directly_ reading the state of the shutdown circuit as opposed to _inferring_ it from somewhere else. That said, if we already have a hardware shutdown sense LED, feel free to ignore; additional data on the output of the BSPD circuit can't hurt either

vehicle/mkvii/software/bspd/bspd.h
```diff
@@ -4,6 +4,9 @@
 #include "libs/timer/api.h"
 
 //////////////////////// VARIABLE DECLARATIONS /////////////////
+uint8_t heartbeat_counter = 0;
```
amit-kumarh: Somewhat of a stylistic choice, but I would define these in your main file, as the header file is more for configuration things in this case. Also, correct me if I'm wrong but it seems like these variables are only being used in one place each? If that is the case they should be `static` variables locally scoped to a single function as opposed to global variables.

[example here](https://www.geeksforgeeks.org/static-variables-in-c/)

amit-kumarh (changes requested):

vehicle/mkvii/software/bspd/bspd.c
H-TejadaDeras: Good point! I will revert the change.

vehicle/mkvii/software/bspd/bspd.c
H-TejadaDeras: I thought it was, in Line 130, this function is called and it is inside the send_can arguments. I think it works that way, correct me if I am wrong!

vehicle/mkvii/software/bspd/bspd.c
H-TejadaDeras: I am confused by this comment, can you elaborate?

amit-kumarh: So, the pcint0_callback() gets the state of BRAKELIGHT_LL, MOTOR_CURRENT_SENSE, BSPD_SHUTDOWN_SENSE. These are digital pins (ie, their values can be either 1 or 0). So a comment stating these are analog would not be correct.

Analog is something like our brake pressure reading, which is not just a binary on or off)

vehicle/mkvii/software/bspd/bspd.c
amit-kumarh: Oops, you're right! I missed that. All good here

PR #322, opened 2025-03-09
added all board files and a software structure 

Eemac:
(NEEDS UPDATES ON THE SOFTWARE SIDE) - let me know what y'all think.


H-TejadaDeras:
Hi Ian,
I like the overall structure of the repo! Just a few things regarding the boards I worked with (and some other things I found).
- When I was modifying the BSPD Schematics, I deleted a 1-2 pages (connector and shtdwn_lsd kicad files). I didn't remove them because I didn't want to break anything, feel free to remove them if they don't break anything!
- When making the LV Motherboard PCB Design, I used a footprint to create the PCB key silkscreen, consider adding that, it is in the Footprints Folder 
- throttle board has ibom included
- Do we want to leave the "sym-lib-table" files (one is found in the Sus. Strain Board and Power Thermistor Board directories)
- I think we should keep the file structure between the hardware and software folders the same (as in we should have the subteam names and then the folders for the software for each individual board just like in the hardware folder)

